PR #20, opened 2022-10-18
Run fontbakery against variable fonts

### LINKED ISSUE #12: Test variable font build with Fontbakery
It appears that our default Fontbakery checks are on the single static instance file?  Can we also add checks on the variable font file generated with the new build workflow?

RickyDaMa:
Will close #12

On a similar note, did we want to proof against variable fonts as well?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ venv/touchfile: requirements.txt
 	touch venv/touchfile
 
 test: venv build.stamp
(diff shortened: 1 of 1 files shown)


chrissimpkins:
> On a similar note, did we want to proof against variable fonts as well?

Yes, I suggest that we do.  But it looks like the proofer builds are down as of the latest pushes in David and Santiago's branch?

chrissimpkins (approved):

RickyDaMa:
> It looks like the proofer builds are down as of the latest pushes in David and Santiago's branch?

If you'd like me to take a look I can do, but open an issue for it please so I can track my findings

PR #35, opened 2022-11-14
Source update 2022-11-14

chrissimpkins:
Source update for review by teams who need to see the latest Latin Roman master / feature architecture and designs.

Taken from the upstream Font Bureau repository at commit https://github.com/FontBureau/fb-GoogleSansFlex/commit/73f70739c6dc4799e3ed9654d171cda7bc484ec7

cc @MariannaPaszkowska @belluzj @RickyDaMa @NeilSureshPatel

Issue #27, opened 2022-11-07
Pull request review helper tools


RickyDaMa:
Reviewing merge requests is currently difficult, among other reasons because it's difficult to know the scope of the changes: which glyphs have changed in which masters. The GitHub native UI doesn't say because it takes too long to load and then you need to review individual XML files to know if the outlines have changed or only an anchor for example. While Font Gardener (#25) aims to be a longer-term solution, there are still useful things we can do in the interim with scripts & GitHub automations.

Ideas:
- [x] #211
- [x] List added/changed/deleted glyphs
- [x] Integrate [ufodiff](https://github.com/source-foundry/ufodiff) (as an implementation of the above?) (#36)
- [x] diffenator2 (#26)
- [x] file size reports (prior work: [report-filesize.py](https://github.com/googlefonts/googlesans/blob/066859050aba6223267d830da98d90b5ede3e979/qa/report-filesize.py)) (#66)

RickyDaMa:
Hi @chrissimpkins & @MariannaPaszkowska, further to the above ideas, I'd appreciate some guidance from yourselves as to how we want to provide these review helpers.

Options:
1. Local scripts. Scripts kept as part of the repository that can be run locally if wanted, takes time and means you have to go out of your way to perform the analysis
2. CI. Scripts or actions are run as part of CI and their output is found amoungst the terminal output within the workflow. Can potentially have artifacts for downloading other formats where needed (e.g. an output HTML file)
3. GitHub bot. Could post results of script/tool runs in nicely formatted markdown directly in PRs. Easy for everyone to use/read, but has the biggest up-front time cost for development. Could potentially be used on other font projects though

Internally we're erring on the side of local scripts as the preference, but I wanted to gather your thoughts before starting work

PR #36, opened 2022-11-15
ufodiff on CI

RickyDaMa:
Uses a docker-based GitHub action to run ufodiff on the repo.

~~Currently configured to compare against master on branches called `import-*` (* being anything)~~ Runs on all pushes to branches except `main`

Keeping as draft until I know it works as intended

### DIFF
--- a/.github/actions/ufodiff/Dockerfile
+++ b/.github/actions/ufodiff/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:3.8.15-alpine
+
+RUN apk add git
+
+RUN pip install ufodiff
+
+COPY entrypoint.sh /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/.github/actions/ufodiff/README.md
+++ b/.github/actions/ufodiff/README.md
@@ -1,0 +1,5 @@
+# ufodiff GitHub action
+
+Runs [ufodiff](https://github.com/source-foundry/ufodiff) in a Docker container so you can make it a GitHub action
+
+Currently only supports the `delta*` subcommands
--- a/.github/actions/ufodiff/action.yml
+++ b/.github/actions/ufodiff/action.yml
@@ -1,0 +1,18 @@
+name: 'ufodiff'
+description: 'A UFO source file diff tool for collaborative typeface development projects'
+inputs:
+  subcommand:
+    description: 'ufodiff delta subcommand'
+    required: true
+    default: 'delta'
+  compare-against:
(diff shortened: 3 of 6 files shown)


RickyDaMa:
Hi Chris & Marianna!

This PR will add a CI step to run [ufodiff](https://github.com/source-foundry/ufodiff) on every push that isn't to `main` and builds successfully. Currently, this is configured to compare against `main`, telling you what glyphs have been added/updated/deleted.

This is achieved using a GitHub action I wrote, which lives in this repository at `.github/actions/ufodiff`. I chose to keep it in this repository as it doesn't really feel suitable for general use (it's not really a tool designed to be used in CI), so there's not much point in making this more open. The GitHub action works by checking out the repository, then ufodiff is run on it (through a Docker container).

I have chosen to create a separate GitHub workflow (note: workflow != action) to run this program in, as part of a to-be-expanded regression testing suite. This makes little difference to how you interact with the workflow, it just means it can be triggered differently to the existing build and release workflow (because I don't want ufodiff running on `main`). The downside here is that I'm not actually able to test that the final state of this PR (as of the time of writing) is configured correctly. That is because GitHub won't let workflows trigger workflows (which is how this new regression suite works) unless they're on the default branch (i.e. `main`) ([citation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run), see 2nd note box). I have tested it through other means and am confident it should work as expected once merged. We would then be able to use it to see what's changed with #35, once that PR builds.

If desired, I can add a means to run this action locally, however I'm not sure there's much of a value proposition to that when you could just use ufodiff directly instead of faffing with Docker. I do already have something mostly jerry-rigged together though, as I needed it for testing.

Apologies for the small essay, interested to hear your thoughts!

(Related issue #27)

chrissimpkins:
ufodiff is unmaintained at this stage.  IMO I would not rely on it in this project unless someone is prepared to maintain it.

RickyDaMa:
For now we'd like to use ufodiff initially as it's a good already-existing starting point (that works currently even if it's not seeing active development/maintenace). We can evaluate later on if the metrics/output it provides is useful on the next source update PR, and if it's something that is a value-add, we can consider taking on the maintenance burden or writing a simplified/bespoke version for GS Flex.

What do you think @chrissimpkins?

chrissimpkins:
Sgtm!

RickyDaMa:

If you guys are happy, you are most welcome to merge whenever :)

MariannaPaszkowska:
Thank you, @RickyDaMa!

Issue #41, opened 2022-11-18
Character set for version 1.000


MariannaPaszkowska:
Here's the proposed character set for version 1.000. Please review and comment. @dberlow 
CC @chrissimpkins @davelab6 

Targeted language coverage:
```
Basque		eus
Catalan		cat
Croatian	hrv
Czech		ces
Danish		dan
Dutch		nld
English		eng
Finnish		fin
French		fra
Galician	glg
German		deu
Hungarian	hun
Icelandic	isl
Irish		gle
Italian		ita
Lithuanian	lit
Latvian         lvs
Malaysian	zlm
Norwegian	nor
Polish		pol
Portuguese	por
Romanian	ron
Slovak		slk
Slovenian	slv
Spanish		spa
Standard Estonian ekk
Standard Latvian  lvs
Swedish		swe
Swiss German	gsw
Turkish		tur
Welsh		cym
```

Targeted feature support for this version is defined in an issue #33

Unicode encoded and basic:

``` 
.notdef None
.null None
uni000D 000D
NULL 0000
space 0020
exclam 0021
quotedbl 0022
numbersign 0023
dollar 0024
cent 00A2
currency 00A4
dollar 0024
Euro 20AC
percent 0025
ampersand 0026
quotesingle 0027
parenleft 0028
parenright 0029
asterisk 002A
plus 002B
comma 002C
hyphen 002D
period 002E
slash 002F
zero 0030
one 0031
two 0032
three 0033
four 0034
five 0035
six 0036
seven 0037
eight 0038
nine 0039
colon 003A
semicolon 003B
less 003C
equal 003D
greater 003E
question 003F
at 0040
A 0041
B 0042
C 0043
D 0044
E 0045
F 0046
G 0047
H 0048
I 0049
J 004A
K 004B
L 004C
M 004D
N 004E
O 004F
P 0050
Q 0051
R 0052
S 0053
T 0054
U 0055
V 0056
W 0057
X 0058
Y 0059
Z 005A
bracketleft 005B
backslash 005C
bracketright 005D
asciicircum 005E
underscore 005F
grave 0060
a 0061
b 0062
c 0063
d 0064
e 0065
f 0066
g 0067
h 0068
i 0069
j 006A
k 006B
l 006C
m 006D
n 006E
o 006F
p 0070
q 0071
r 0072
s 0073
t 0074
u 0075
v 0076
w 0077
x 0078
y 0079
z 007A
braceleft 007B
bar 007C
[...]

PR #42, opened 2022-11-21
Automated GitHub workflow import process

### LINKED ISSUE #28: Set up import process for vendor sources
Writing out the rough import process as exists in Google Sans ([citation](https://github.com/googlefonts/googlesans/blob/35577ff11a979c7c0d772d7cec75f3ae9cb00f43/docs/MAINTAINER.md#steps-in-common-for-designspace--ufo-or-glyphsapp-imports)), complete with check boxes for what is available in this repo on an automated basis:

1. [x] Run the import script on the designspaces, e.g. `python3 scripts/gs-merge-designspace.py --source source/GoogleSans/staging/GoogleSansSomeScript.designspace --target source/GoogleSans/GoogleSans.designspace --import-glyphs-file source/GoogleSans/staging/import_glyphs.txt`
    * Script looks general enough to be useable for Google Sans Flex
    * Locations for staging etc. ought to be standardised to allow for automation. Although, we may be able to do everything in-repo with [git worktree](https://git-scm.com/docs/git-worktree) or another less elegant solution
    * [Script](https://github.com/googlefonts/googlesans/blob/main/scripts/gs-merge-designspace.py)
3. [ ] Extract and merge feature code
    * Difficult to automate but unlikely to be an issue for Jan release due to simple glyphset
    * Instead focus on notifying if the feature code has changed
    * No need to merge until Jan on first release, so can blindly copy for now
2. [ ] Run anchor propogation on glyphs that need it
3. [x] Normalise the designspace
    - Script already exists for this
    - [Script](https://github.com/googlefonts/googlesans/blob/main/scripts/gs-normalize-designspace.py)

RickyDaMa:
Will close #28 

Outline of steps that will be taken:

1. You trigger the workflow over in [the Actions tab](https://github.com/googlefonts/googlesans-flex/actions) (there'll be an entry on the left labelled "Import sources to main")
2. You'll be asked to enter a branch name

The CI then:

1. Checks out the main branch and the user's requested branch using `git worktree`
2. These and the import scripts are then mounted into a Docker container, which will have Python & the scripts' dependencies nicely sandboxed
3. Docker container runs the import scripts
4. CI then creates a new branch called import-<branch> (whatever was given in 2.), commits, and opens a PR

Additional design considerations / unsolved problems:

- [x] Updating an existing open branch/PR for an import - should be possible with conditionals
- [x] Ensuring you can run the whole process locally if you want. Differences in environment in local & GH action make this challenging


RickyDaMa:
Limitations/Considerations with the current process:
- Glyph list omissions are usually fatal errors, but have been downgraded to warnings for a friendly experience. Being tracked in #41. The glyph list that we anticipate using is in `.github/actions/import/glyph-list.txt`. Please review and/or make sure it's kept up-to-date
- The import scripts now allow for missing lib keys, e.g. `public.skipExportGlyphs` & `public.postscriptNames` (the latter of which used to be a fatal error
- New UFOs aren't touched currently. This is challenging to automate as it requires modifying the Designspace, and it can be harder to tell if the changes are 'correct' or not
- You can't run the local version from the Makefile, because make is not designed to support positional arguments. [Justfiles](https://github.com/casey/just) could be a solution to this

Still to-do at time of writing:
- [x] Port local script to GitHub workflow
- [x] Allow for updating an existing import branch

---

In its current state, this PR is usable as an MVP in my opinion

You are able to run the automated import script locally given the following prerequisites:
* Linux/Mac/WSL bash shell
* Docker
* SSH GitHub authentication

There's more documentation in the repo, but in short, you just run `./.github/actions/import/local.sh <branch-to-import>`

madig:
@MariannaPaszkowska regarding UFOs, can you tell us whether you'd expect UFOs to change position in the Designspace or new ones coming in or some being removed before the 1.000 release?

MariannaPaszkowska:
@madig New sources will be added to support the GRAD axis. Existing sources are in active development, so I think it's possible that some UFOs might change position in the Designspace. Let me confirm all that with @dberlow 

CC @chrissimpkins

madig:
Ah. Do you expect new script additions to have varying support for axes, like GS, where plently of scripts have no GRAD masters?

MariannaPaszkowska:
Yes, in further stages, some scripts might have custom axes local to the script.

RickyDaMa:
Happy Friday! Lots more functionality just pushed 😄 

What you can now do:
- Start an import locally
- Start an import from GitHub ✨
- Update an import from GitHub ✨

To do:
- [x] Update an import locally
- [x] Document how to run the workflow (probably in the main README)

I'll demo everything on Monday :)

There are limitations of the automatic process due to uncertainty in how designspaces/masters may change, but I propose we create a follow-up issue after this is merged to discuss these further and refine the Pythonic merging. Most of what I've done here is just automating, nothin novel in terms of merging strategy

RickyDaMa:
Everything done! You can now start and update imports both locally and through GitHub. Documentation updated

Ready for merge whenever 😄

RickyDaMa:
How it looks:

image

image

(wait ~2 minutes)

image

image

Updating an existing PR is very similar, just a new commit gets tacked onto the existing branch

RickyDaMa:
(CI failures appear to be due to downtime of a build dependency mirror, completely unrelated to this PR)

.github/actions/import/scripts/gs-merge-designspace.py
```diff
@@ -0,0 +1,379 @@
...
+
+"""Merge an incoming Designspace into an existing Designspace.
+
+This script will import glyphs and groups specified in import text files
+(one name per line) and kerning pairs that mention either of them. It will
+also update each UFO's public.glyphOrder and public.postscriptNames lib keys
+with entries for all imported glyphs, as well as public.skipExportGlyphs in
+Designspace and UFOs.
+
+It does not import any font info, global or local guidelines or features.
```
MariannaPaszkowska, line 24: Minor improvement suggestion: Can we make this comment more specific? Which font info won't be imported?

RickyDaMa: @madig

.github/actions/import/scripts/internal/gdef.py
```diff
@@ -0,0 +1,322 @@
...
+
+# pyright: basic
+
+from typing import Dict, Optional
+
+import glyphsLib.builder.constants
+import glyphsLib.glyphdata
+import ufoLib2
+
+KNOWN_BASES = {
```
MariannaPaszkowska, line 29: Are Devanagari and Khmer glyphs mentioned here on purpose?

madig: Yes, because I needed to do it for import work in GS. I copy-pasted the file over.

MariannaPaszkowska (approved):

madig:
Don't merge yet, I need to figure out an import pecularity. Will post a changelog after I figured it out.

madig:
A bare fontmake works, but using `gftools build` messes up the axis names. Weird. I'll look into it.

Oh, btw, do you want us to preserve the folder structure? I see roman and italic subfolders in FB's branch.

MariannaPaszkowska:
@madig folder structure as requested by FB:

/Main
  /FB-branch
     /Fonts, contains latest builds
       /Previous, contain previous builds
     /sources contains only folders
      /legacy contains sources and fonts,
       /ProductSans,
       /GoogleSans and
       /GoogleSans variable
      /experiments contains drawings requested and glyphs not yet included inLatest.
        /figures
        /features
        /misc
      /regular contains active sources and ds files
        /ROND
      /italic  contains active sources and ds files
        /ROND

CC @dberlow

madig:
Ok. Also also, have you decided on STAT entries and instance locations yet?

madig:
Ok, I think this is actually mergable now. The folder structure should be changed in a different PR.

.github/actions/import/scripts/gs-merge-designspace.py
madig: Any font info. The fontinfo.plist file isn't touched.

Unless we're replacing the Designspace, then it is imported wholesale from the font to be imported.

madig:
Specifically, I changed the import scripts to

1. Drop the group import parameter, as it has been unused for a while; glyph import lists are enough.
2. Add `--replace-target-designspace` and `--follow-glyphs`, which allows for an incomplete Designspace in flux
3. Makes missing production names an info rather than warning, because we can control them in a different way.

madig:
@MariannaPaszkowska @chrissimpkins please review. The FB location update necessitates a separate PR to move the current sources into the expected location and make the scripting here work.

sources/config.yaml
```diff
@@ -14,3 +14,141 @@ buildOTF: false
...
+    value: 125.0
+  - flags: 0
+    name: ExtraExpanded
+    value: 151.0
+- name: Optical size
+  tag: opsz
+  values:
+  - flags: 0
+    name: 6pt
+    value: 6.0
```
MariannaPaszkowska, line 85: I think we should be using ranges; let's discuss that. It can be addressed after the merge.

PR #44, opened 2022-11-21
Update dependencies

### LINKED ISSUE #75: `make update` command broken
The Makefile contains

https://github.com/googlefonts/googlefonts-project-template/blob/d66be4c457b4974fb59e575b0b1f38912dc5becf/Makefile#L52

but the `$(dependency)` variable is undefined. Running pip then crashes (after activating the venv, which isn't done like in the build command):

```
pip install --upgrade ; pip freeze > requirements.txt
ERROR: You must give at least one requirement to install (see "pip help install")
```

I also find the in-place updating of requirements a bit unfortunate. It perpetuates dependencies even if they're not used anymore because the primary/transient distinction gets lost. A cleaner way would be to have a file requirements.in:

```
fontmake>=2.4
fontbakery>=0.8
gftools[qa]>=0.7
drawbot-skia>=0.4.8
sh>=1.14.2
bumpfontversion>=0.2.0
```

and use `pip-compile requirements.in` (https://github.com/jazzband/pip-tools) to generate the actual requirements.txt with all pinned deps.

madig:
This fixes https://github.com/googlefonts/googlefonts-project-template/issues/75 and updates dependencies.


madig:
@chrissimpkins @MariannaPaszkowska this is ready to go in. This is a generic update-everything PR while we are still in motion.

chrissimpkins (approved):
Recommend merging this before we get heavily into QA, particularly the visual QA.  Ideally we firm up these requirements and then not change them again in advance of v1.0.  Are there any other expected dependency updates that will be required between now and the January release?

madig:
No, barring bugs that need fixing. The ufo2ft kerning splitter will only become relevant when there's more scripts and kerning I suppose.

MariannaPaszkowska:
Thank you, @madig!

PR #60, opened 2022-11-29
Fix ufodiff not running / Better CI triggers

RickyDaMa:
This change will ensure ufodiff runs. Conditional workflows seem to temperamental, so I've fallen back on running every push (not to `main`)

The regression pipeline is skipped when nothing font-related has changed. I've also made the main build be skipped when READMEs or GitHub actions/workflows are changed. Might as well save costs where we can

### DIFF
--- a/.github/actions/ufodiff/Dockerfile
+++ b/.github/actions/ufodiff/Dockerfile
@@ -1,6 +1,6 @@
-FROM python:3.8.15-alpine
+FROM python:3.8.15-slim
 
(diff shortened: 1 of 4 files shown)


RickyDaMa:
@MariannaPaszkowska apologies for the unexpected delay on this one, everything should now be ready to merge!

MariannaPaszkowska:
Thank you, @RickyDaMa!

PR #66, opened 2022-12-05
File size analysis on CI

RickyDaMa:
Part of #27 

Runs Chris' beloved [file size report script](https://github.com/googlefonts/googlesans/blob/35577ff11a979c7c0d772d7cec75f3ae9cb00f43/qa/report-filesize.py) from Google Sans on the build artifacts.

This way you don't have to run the tool locally (though I'm planning on adding support for doing so), and the information is included in every build

Leaving as draft while I validate the setup works and get the local script up & running


RickyDaMa:
image

RickyDaMa:
Hi @MariannaPaszkowska, this PR is ready for review/merging whenever you have time! Not a terribly urgent one, but another tool in your PR review toolkit 😃

Makefile
```diff
@@ -28,7 +28,7 @@ venv/touchfile: requirements.txt
 	touch venv/touchfile
 
 test: venv build.stamp
-	mkdir -p out/ out/fontbakery
+	mkdir -p out/fontbakery
```
MariannaPaszkowska, line 31: @RickyDaMa is this change intentional?

RickyDaMa: Hi Marianna, yes it is. `mkdir -p` creates all intermediary folders itself, so there's no need to specify `out` and `out/fontbakery` separately.

A lot of people use `-p` as "don't error if it already exists", when based on the flag's full name that's not even it's primary intention!

image

You can also be assured this isn't causing an issue given the build pipeline passed on this PR

PR #67, opened 2022-12-05
Add scripts to more easily manage glyph metadata

### LINKED ISSUE #117: Glyph names in sources - automatic renaming process
Let's create an automatic renaming process to make the glyph names from the FB sources consistent with https://github.com/googlefonts/googlesans-flex/blob/main/.github/actions/import/glyph-list.txt

Font Bakery fail: [Glyph names are all valid?](https://github.com/googlefonts/googlesans-flex/actions/runs/3707745278/jobs/6284525053#step:9:284)
@belluzj @madig 
CC @chrissimpkins

madig:
To discuss further.

This PR adds a script to extract, regenerate and apply glyph metadata from and to UFOs, using my tool at https://github.com/daltonmaag/ufo-glyphdata-manager to do the heavy lifting. It also changes the importer to not import production names, so that the CSV file remains the authority.

This goes some way towards centralizing information like FontGardener is planned to do, but with UFOs. It makes it easier to check and edit information in one CSV file and then apply it to all UFOs uniformly. Information is left in the UFOs currently, but could be deleted and stored only in the CSV, putting it back before the sources hit the compiler.


RickyDaMa:
Sample CSV so you can see what sort of metadata is extracted/stored

[glyph_metadata.csv](https://github.com/googlefonts/googlesans-flex/files/10288050/glyph_metadata.csv)

belluzj:
It would be good to write in the MAINTAINER doc what "extract", "update", "apply" mean and when they're useful

belluzj:
This PR is no longer relevant, it was more targeted at the V1 situation where we wanted to only import outlines from the vendors and keep control over the engineering; we've now moved to trusting Octavio's team with as much as possible, so we don't need this anymore.

PR #68, opened 2022-12-05
Fix import workflow

RickyDaMa:
Correct the path to the glyph list & change some references to the DaMa fork of Google Sans Flex

Apologies these weren't caught in the first place


MariannaPaszkowska:
Np, thank you, @RickyDaMa!

PR #70, opened 2022-12-06
Import automation fix: update glyphsLib to latest

RickyDaMa:
Following #68, the [latest error](https://github.com/googlefonts/googlesans-flex/actions/runs/3628450865/jobs/6119483529#step:4:119) in the workflow appears to be because the script relies on a feature only added in the latest version of `glyphsLib`. I've updated the dependency pulled in so that this should work going forwards


MariannaPaszkowska:
Thank you, @RickyDaMa!

Issue #73, opened 2022-12-07
Create a ufoLib script for an automated burndown chart


MariannaPaszkowska:
User story: The user gets a visual report of the progress of the font project at every new font build. It should be easy to understand which and how man glyphs are in-progress or finished. It should be in an easy-to-use format (could be a PDF).

The script will pull the UFO '[Mark Colors](https://robofont.com/documentation/tutorials/using-mark-colors/)' marks for color coding glyphs from editor sources and will report on progress using that data. 

@JeremieHornus and @justvanrossum created something like that for the GS CJK build CI (RoboCJK). Perhaps they could share their experience and offer advice? :)

CC @davelab6 @chrissimpkins

belluzj:
Idea for implementation:
* target = when the project is done = sum(number of glyphs in each source)
* current progress for a given commit = sum(number of glyphs marked "green" in each source)

Then a script could check out each commit in turn, compute the current progress and plot a chart like so:
* x axis = time of commit
* y axis = target - current progress for the commit

The resulting curve should be going downwards and hit 0 when all glyphs are green.

@MariannaPaszkowska Is this what you have in mind?

JeremieHornus:
Curve going downwards or upwards with the sum of glyphs actually finished could either work.
What we did previously was to include the deadline in the chart with a straight line from the begining to the goal so that the team knows if late or ahead of schedule.

JeremieHornus:
we also included several 'status' colours (3 to 5 depending on projects) to be able to differentiate glyphs in-progress to finished ones.

belluzj:
Sounds good, and the increases in status would count towards the goal? Like:

- target = when the project is done = sum(3 * number of glyphs in each source)
- current progress for a given commit = sum(3 if green, 2 if yellow, 1 if orange, 0 if red or no color for each glyph in each source)

MariannaPaszkowska:
Thanks, @JeremieHornus! What was the color coding? Did you distinguish between different axes?

@belluzj What Jérémie is describing is what I think we need. We can ignore reporting for the specific glyphs for now. Can you extract deadlines from the milestones?

JeremieHornus:
Color coding was:
- Red: started / in progress
- Orange: check 1
- Yellow: check 2
- Blue: check 3
- Green: validated

We distinguish each source, once a source is Green there is a lock mechanism that prevents further modification.

A glyph is considered finished (validated) if all its sources are Green.

belluzj:
[...]

PR #76, opened 2022-12-08
Progress burndown chart

belluzj:
Implements #73.

Todo: 
* [ ] Better define the expected numbers of UFOs for the upcoming milestones (to have more accurate slanted lines)
* [ ] Integrate into a GitHub workflow


belluzj:
Here is a preview of the output of the script, on a test branch that I prepared with fake progress:

Roman_-_version_1 000_counts
Roman_-_version_1 000_progress

The colored area in the middle is the range of commits that have mark colors. In the first graph, all glyph statuses are given equal importance, and in theory if all the UFOs exist and all contain the target glyph set, the colored area should be as tall as the blue line (except in case of sparse sources; to be refined if that's the case). It's up to the reader of the graph to interpret that yellow glyphs are only halfway done.

The second graph is scaling the glyph statuses by a "progress percentage", so green glyphs = done are shown full height, yellow glyphs = 50% done are shown half-height, and not started yet glyphs (red) are not shown at all.

The script takes in [a configuration object](https://github.com/googlefonts/googlesans-flex/pull/76/files#diff-e16709e3e926d793c7c80e42c2575346ea1b42bbfda15bb96b52468cadd98ed5R69-R110) that defines:

* which range of commits to analyze
* for each milestone:
    * start and due dates, to draw the blue line
    * target number of glyphs and UFOs, to set the top-right corner of the blue line (glyphs * UFOs)
    * a filter to know in which UFOs to look for colors (e.g. to make a milestone only about uprights)
    * a list of possible glyph statuses, with the mark color to use for matching them.
 
We can adjust the configuration when once we know the exact mark colors that the FB team will be using to annotate progress; for now I've used retina-burning pure RGB colors.

Let us know what you think.

belluzj:
Here is the current result on fb-wip:
Roman_-_version_1 000_counts

I have solved the question of the exact mark colors by using the "colorsys" python module from the standard library to extract the hue of the mark colors and categorize the ones currently in use:
image


I have a new question now related to the target amount of glyph instances:
* either we could "predict" the target amount by computing number of glyphs * number of UFOs
  * doesn't work for sparse sources (they have fewer glyphs than a full source)
  * is the final number of UFOs already decided?
* or we could measure what we find on the branch, but it makes the target "move" with the state of the branch.

belluzj:
For now:

- [ ] ignore sparse sources
- [x] put all milestones on the same graph

For the future:
* differentiate composite vs drawn, assign smaller weight to composites in the diagram
* try to speed up

belluzj:
@MariannaPaszkowska @chrissimpkins I have a question about the target number of main sources for the v1.000 and v1.100 milestones: do you know how many UFOs will be used to support the GRAD axis? Will GRAD-50 and GRAD50 variant be drawn for all GRAD0 sources, or will only some of the GRAD0 sources get matching GRAD-50 and GRAD50? If so, do you know which? Thanks in advance!

belluzj:
I have improved the output a bit, see below:

gs-flex-progress


The targets for each milestone are still not clear to me, in terms of how many UFOs are planned in total.

Next we could generate this graph as part of the push workflow to fb-wip? Or maybe with a manual trigger, if you don't want to generate it on each push? The script currently takes 2 minutes 30 to run on my machine.

@MariannaPaszkowska @chrissimpkins let us know what you would prefer.

MariannaPaszkowska:
@belluzj January release is planned to support all axes: wght, wdth, ROND and GRAD (and only Roman). I expect 120 basic masters (+some sparse masters). 

Let's generate this graph on each push.

belluzj:
Updated with the expectation of 120 UFOs for v1.000 (and 120*2 for v1.100, because of the italic):

gs-flex-progress

@madig will automate this as part of a GitHub action on fb-wip.

madig:
I added a workflow, but it does not seem to trigger. Maybe it needs to be in main first? @chrissimpkins @MariannaPaszkowska I suppose this is ready for review and merging?

MariannaPaszkowska:
Thank you @belluzj, @madig!

PR #80, opened 2022-12-09
Simplify workflow trigger

RickyDaMa:
Hopefully this will get the workflow to run on #78 

The `paths-ignore` shouldn't be stopping the build workflow from triggered, but we have no idea what else it could be so this is our debugging step

@MariannaPaszkowska sorry to bother you with this one

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,12 +1,6 @@
 name: Build font and specimen
 
-on:
-  push:
-    paths-ignore:
-      - .github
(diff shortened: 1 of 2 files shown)


RickyDaMa:
Upon further reading, looks like there's a limit on the number of files GitHub will check have changed before deciding to run the workflow (like 300). Considering UFO diffs are thousands of files it's probably not a great combination

MariannaPaszkowska:
@RickyDaMa Thank you!

Issue #31, opened 2022-11-07
Automation for interpolation artifacts


RickyDaMa:
Because the designspace is so big with so many axes, some amount of computer help is going to be needed for this.

It may helps to have list(s) of designspace locations that need reviewing. Exhaustive testing will scale poorly as more axes are added.

Ideas *(tags don't indicate responsibility, just whose idea it was)*:
- video diffenator (@m4rc1e)
- Fontbakery outline tests (these rely on heuristics and thus aren't foolproof) (@madig)
- checking for smooth/harmonious nodes across the designspace (@belluzj)
- automated specimen generation (some prior work already done) (@chrissimpkins)

belluzj:
We can start looking into this as well.

belluzj:
The point is to support reviewers to know where to look for the problems (find the glyph + designspace location)

MariannaPaszkowska:
Ideas from this issue were out into separate issues.

PR #102, opened 2022-12-12
Initial draft of a statistical outlier finder

madig:
Related: https://github.com/googlefonts/googlesans-flex/issues/31

This script finds points in glyph instances that somehow are a statistical outlier. At the moment, it looks at the angle at which a node is bent, and if e.g. a node is at a 90° angle everywhere, except at one location where it is at 180°, then there may or may not be an interpolation issue around that location.

Example: asas

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,3 +49,5 @@ A script used to find angles (between the incoming and outgoing handles or point
 3. In a separate terminal or tab, go to the root directory of GSF and activate the venv (e.g. run `make venv` and then `source venv/bin/activate`).
 4. Run the script like this: `python scripts/find-problems.py ../googlesans-flex-fb-wip/sources/roman/GoogleSansFlex.designspace`. The Designspace can be anywhere, but the script should be run from your venv.
 5. The script will create a CSV file where it lists glyph points and their locations where the angle of the incoming and outgoing handle or on-curve deviates from other masters. It gives you a link to click, which brings you to the glyph in Fontra and marks the offending point for you. Wiggle the location sliders to see if there is a problem or not.
+
+You can also inspect variable TTFs. Note that the locations might be slightly off then.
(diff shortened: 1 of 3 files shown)


belluzj:
@MariannaPaszkowska Let's discuss in our meeting tomorrow. We could have a call with you to show you how to use the script and look at the results; sometimes the glyphs that are spotted don't have any problem at all, sometimes there is something happening and it's hidden in an overlap or inside a stroke, so it's not an issue either, and finally sometimes it does spot actual interpolation bugs.

belluzj:
TODO: better UX
* make it work with videoproof
* output a spreadsheet
* add some visual clues for where to look, e.g. arrows to point at points
* improve false positives

belluzj:
New test:
* point type (smooth or sharp)

madig:
So, the current find-problems.py script is used like this:

1. First, `git clone https://github.com/BlackFoundryCom/fontra` somewhere and set it up inside a venv as described in [the Readme](https://github.com/BlackFoundryCom/fontra#install-from-the-source-code).
2. Then, start it by pointing it to the directory containing GoogleSansFlex.designspace, like so:
    * `fontra --launch filesystem ../googlesans-flex/sources/regular/`
3. In a separate terminal or tab, go to the root directory of GSF and activate the venv (e.g. run `make venv` and then `source venv/bin/activate`).
4. Run the script like this: `python scripts/find-problems.py ../googlesans-flex-fb-wip/sources/roman/GoogleSansFlex.designspace`. The Designspace can be anywhere, but the script should be run from your venv.
5. The script will create a CSV file where it lists glyph points and their locations where the angle of the incoming and outgoing handle or on-curve deviates from other masters. It gives you a link to click, which brings you to the glyph in Fontra and marks the offending point for you. Wiggle the location sliders to see if there is a problem or not.

madig:
@MariannaPaszkowska This is ready to merge for now. Further work can be done in additional PRs. I added the above comment to the Readme.

PR #104, opened 2022-12-13
Import workflow: import designspaces based on build configuration

RickyDaMa:
The PR title is admittedly a bit of a word salad. Essentially, the import workflow used to have hardcoded the locations of the designspace files. Now, it reads them from `sources/config.yaml`, which is the file the build pipeline uses to decide which fonts to build. It made sense to make the import workflow use this too

Technical jargon for people code-reviewing:

The result of this was rewriting the entrypoint of the import action in Python. This gave easy access to YAML parsing, whereas in shell script I'd've been doing certifiably sketchy things to get the information. I've also updated the merge and normalise scripts so they can more easily be run programmatically or by commandline, by parameterising their `main` functions and doing argument parsing in `if __name__ == "__main__"`.

To be able to test this functionality, I updated the `local.sh` script to be compatible with the changes to the import action inputs/mounts (from #68). I also added licence blurbs to a couple of shell scripts that didn't already have them


.github/actions/import/entrypoint.py
```diff
@@ -0,0 +1,52 @@
...
+from scripts.gs_merge_designspace import main as merge
+from scripts.gs_normalize_designspace import main as normalize
+
+REPLACE_TARGET_DESIGNSPACE = True
+FOLLOW_GLYPHS = True  # while sources are in flux
+
+logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+
+# Takes designspaces specified in the sources/config.yaml of the SOURCE branch
+with open("/github/workspace/source/sources/config.yaml", "r") as file:
```
madig: You can also do this with less indentation by doing `yaml_txt = Path(...).read_text()`.

.github/actions/import/entrypoint.py
```diff
@@ -0,0 +1,52 @@
...
+    gftools_config = yaml.safe_load(file)
+    for designspace_path in gftools_config["sources"]:
+        logging.info(f"Merging {designspace_path}")
+        src_ds_path = designspace_path.replace("regular", "roman", 1)
+        merge(
+            Path(f"/github/workspace/source/sources/{src_ds_path}"),
```
madig: This is fine, but you can also do things like 

```py
WORKSPACE_DIR = Path("/github/workspace/")
SOURCE_DIR = WORKSPACE_DIR / "source"
```

etc., just FYI.

madig (approved):

RickyDaMa:
Updated the PR to include #147. Made changes to the script consistent with how I adjusted the others (parameterised main, made argument parsing only happen in the commandline, renamed to snake case), see https://github.com/googlefonts/googlesans-flex/pull/104/commits/62849e8e568927bcaa733ee1147f849e8d2267f0

RickyDaMa:
@MariannaPaszkowska note that this branch now contains an important fix (https://github.com/googlefonts/googlesans-flex/pull/104/commits/d01afb4720fe343becb73aa3b50b8f12f0fe888d) that ensures that the fontbakery checks run on new PRs are using the checks on the `main` branch, instead of `fb-wip`, which are very outdated. Increasing the priority of this PR

madig (approved):

.github/actions/import/scripts/gs_merge_designspace.py
```diff
@@ -46,64 +46,28 @@
...
-        for name in parsed_args.import_glyphs_file.read_text().split("\n")
-        if name
+        name.strip() for name in import_glyphs_file.read_text().split("\n") if name
     ]
     import_glyphs = set(import_glyphs_verbatim)
 
-    replace_target_designspace: bool = parsed_args.replace_target_designspace
-    follow_glyphs: bool = parsed_args.follow_glyphs
+    replace_target_designspace: bool = replace_target_designspace
+    follow_glyphs: bool = follow_glyphs
```
belluzj, line 63: Not needed

.github/actions/import/scripts/gs_merge_designspace.py
```diff
@@ -379,4 +343,49 @@ def find_matching_source(
...
+    )
+    parser.add_argument(
+        "--follow-glyphs",
+        action="store_true",
+        help=(
+            "Also import glyphs that are referenced but not explicitly listed in "
+            "the import list."
+        ),
+    )
+    parsed_args = parser.parse_args()
```
belluzj, line 384: This should be the main function, so that (if we had unit tests) a unit test could call `main(["--source", "whatever", etc)` to test the argument parsing.

.github/actions/import/scripts/gs_merge_designspace.py
```diff
@@ -379,4 +343,49 @@ def find_matching_source(
...
+    main(
+        parsed_args.source,
+        parsed_args.target,
+        parsed_args.import_glyphs_file,
+        parsed_args.replace_target_designspace,
+        parsed_args.follow_glyphs,
+    )
```
belluzj, line 391: The separation of this into another function is good; but I wouldn't call it main, something like "merge_designspace()".

.github/actions/import/scripts/gs_normalize_designspace.py
```diff
@@ -47,4 +38,12 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    parser = argparse.ArgumentParser()
```
belluzj, line 41: Same idea here, main() is all the code of the script, have a separate function if you want but call it with a descriptive name like `normalize_designspace()`

belluzj (commented):
Looks good to me apart from small fixes to scripts that can be both run and imported

PR #105, opened 2022-12-13
Add build to import process

RickyDaMa:
A glorified copy-paste job of the `build.yaml` workflow, with some parts removed and paths changed

Currently, file size reporting is broken as it is hardcoded to expect different paths. I will work on this tomorrow

The font artifacts, test results, etc. are instead stored on the import workflow, as opposed to the previous separate setup

@MariannaPaszkowska you can either approve & merge this as-is or you can wait for further fixes for file-size. You can still use this version of the workflow before it's merged by changing the "Use workflow from" branch to be `import-not-triggering-build` when kicking off an import

### DIFF
--- a/.github/workflows/import.yaml
+++ b/.github/workflows/import.yaml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   import-branch:
-    runs-on: ubuntu-latest
+    runs-on: [linux, googlefonts-64cores-256GB]
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


chrissimpkins:
Thank you very much Ricky!  I will be the one pushing source updates as work wraps up by Font Bureau each day.  I am going to merge this PR so that we can test that it functions with tonight's set of updates.  Marianna merged #78 and we will open a new PR to support David and Marianna's review later this evening my time (I'm on EST)

PR #106, opened 2022-12-13
Tune checks

madig:
This changes the tests as discussed with @MariannaPaszkowska.

* Exclude `com.google.fonts/check/vertical_metrics` and `com.google.fonts/check/varfont/regular_opsz_coord` because we use GS as the reference.
* Exclude `com.google.fonts/check/glyph_coverage` because we have our own target
* Exclude `com.google.fonts/check/file_size` because we're past the 1MB anyway and know it.
* Exclude `com.google.fonts/check/font_names` because we want our own naming.
* Make wdth=150 ExtraExpanded instead of 151
* Add dummy STAT entries for GRAD and ROND, as expected by the axis registry
* Use Google Fonts' `com.google.fonts/check/family/win_ascent_and_descent` to check win metrics.


madig:
@chrissimpkins @MariannaPaszkowska I think this is ready for review.

PR #124, opened 2022-12-19
Fix syntax error in burndown workflow

madig:


madig:
Apparently there is more to fix than a syntax error. Sigh.

madig:
@MariannaPaszkowska ready to merge :) You can get a burndown chart from here for this PR: https://github.com/googlefonts/googlesans-flex/actions/runs/3740377042

PR #129, opened 2022-12-20
Metadata configuration script

madig:
Add a script that modifies sources to implement 

* https://github.com/googlefonts/googlesans-flex/issues/125
* https://github.com/googlefonts/googlesans-flex/issues/107
* https://github.com/googlefonts/googlesans-flex/issues/120

PR #147, opened 2023-01-02
Move metadata fixer into import process

madig:
As per https://github.com/googlefonts/googlesans-flex/pull/129#issuecomment-1367439570.


madig:
Tested on our playground fork and it appears to work. @MariannaPaszkowska ready to merge. This will automatically apply various fixups on import, overwriting what is there. That way, FB can focus on more important tasks.

PR #163, opened 2023-01-04
Remove config file copying during import

RickyDaMa:
We should assume that the target (main) has a better config as it's unlikely to be directly updated by Font Bureau

Thanks to @madig for spotting this flaw in #158


RickyDaMa:
Looks like @madig beat me to it! Closing since #162 has been merged

PR #201, opened 2023-01-09
Release workflow: fix bumping version

### LINKED ISSUE #192: Test git tag push release workflow
Confirm that the push of a git tag formatted as `vX.XXX*` prepares a GitHub release with the appropriate release version, and a zip archive with the release files that includes the release version in the zip archive file path.

RickyDaMa:
Will close #192 

It's a Python script for bumping the font version. The old bash / in workflow version was unreliable (venv wasn't yet necessarily made) and pulled in a heavy dependecy (`yq`, a snap). This simple script aims to replace both

I've also pinned the Python version of the CI to 3.11 both for consistency and so we don't get thrust into 3.12 without warning, as Nikolaus is concerned with dependencies when that happens

Technical jargon:

I have to `subprocess.run` `bumpfontversion` because it's not designed to be run in a script as far as I can tell

~~Aside: `bumpfontversion` has no license, and is therefore proprietary (though [source available](https://github.com/simoncozens/bumpfontversion))~~ (edit: it's in the README of the repo)

### DIFF
--- a/scripts/bump-to-tag.py
+++ b/scripts/bump-to-tag.py
@@ -17,6 +17,7 @@ import os
 import yaml
 from pathlib import Path
 
+
 def main():
     try:
         tag_name = os.environ["GITHUB_REF"].replace("refs/tags/", "")
@@ -32,7 +33,7 @@ def main():
     for source in gftools_config["sources"]:
         print(f"Bumping {source} to {tag_name}")
         cmd = subprocess.run(
-            f". venv/bin/activate && bumpfontversion sources/{source} --new-version {tag_name}",
+            f"bumpfontversion sources/{source} --new-version {tag_name}",
             shell=True,
             capture_output=True,
             text=True,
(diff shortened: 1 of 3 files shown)


scripts/bump-to-tag.py
```diff
@@ -0,0 +1,48 @@
...
+        file_text = CONFIG_FILE.read_text()
+    except (KeyError, FileNotFoundError):
+        print("This script is intended to be run by CI during a release. Exiting")
+        exit(2)
+
+    gftools_config = yaml.safe_load(file_text)
+    for source in gftools_config["sources"]:
+        print(f"Bumping {source} to {tag_name}")
+        cmd = subprocess.run(
+            f". venv/bin/activate && bumpfontversion sources/{source} --new-version {tag_name}",
```
madig: I think the activation here may not be neccessary if it's done in the Makefile anyway?

RickyDaMa: I will test this in playground, though I suspect you're right

RickyDaMa: Activation was not necessary, but `shell=True` still is

RickyDaMa:
@MariannaPaszkowska this is ready to merge now. Means that creating a release on GitHub will automatically attach the font archive to the release

PR #213, opened 2023-01-09
Recompute win metrics on import

madig:
This will pick up all vertical metrics changes on import.

@MariannaPaszkowska

### DIFF
--- a/.github/actions/import/scripts/fix_metadata_in_sources.py
+++ b/.github/actions/import/scripts/fix_metadata_in_sources.py
@@ -18,7 +18,6 @@ from pathlib import Path
 
 from ufoLib2 import Font
 from fontTools.designspaceLib import DesignSpaceDocument
-from typing import Optional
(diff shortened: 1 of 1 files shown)


.github/actions/import/scripts/fix_metadata_in_sources.py
```diff
@@ -68,9 +72,35 @@ def main(designspace_path: Path):
...
+
+    for ufo in unique_ufos:
         ufo.save()
 
 
+def recompute_win_metrics(fonts: list[Font]) -> None:
+    ascent = 0
+    descent = 0
+
+    for font in fonts:
```
chrissimpkins, line 117: Are the values defined in this loop consistent across all masters?

madig: We measure all fonts and then write the same values to all in a second pass.

PR #214, opened 2023-01-09
Import process: add/fix production names where needed

### LINKED ISSUE #117: Glyph names in sources - automatic renaming process

RickyDaMa:
Will close #117

If there is an existing postscript name, checks it doesn't have hyphens and removes them if so

If there isn't an existing postscript name, checks the design name doesn't have hyphens and adds a postscript name with none if so


RickyDaMa:
Thought: should I just include this in the fix metadata script? I guess postscript names are metadata given they're a lib key

.github/actions/import/scripts/fix_glyph_production_names.py
```diff
@@ -0,0 +1,51 @@
...
+
+ROOT_DIR = Path(__file__).parent.parent
+POSTSCRIPT_NAMES = "public.postscriptNames"
+
+
+def fix_production_names(designspace_path: Path):
+    designspace = DesignSpaceDocument.fromfile(designspace_path)
+    ufos: list[Font] = designspace.loadSourceFonts(Font.open)
+    for ufo in ufos:
+        production_names: Optional[dict[str, str]] = ufo.lib.get(POSTSCRIPT_NAMES)
```
madig: Best to take prodnames from `designspace.findDefault().font`.

.github/actions/import/scripts/fix_glyph_production_names.py
```diff
@@ -0,0 +1,51 @@
...
+        for glyph in ufo:
+            if not glyph.name:
+                continue
+
+            has_production_name = production_names is not None and glyph.name in production_names
+            if has_production_name and "-" in production_names[glyph.name]:
+                existing_production_name = production_names[glyph.name]
```
madig:
```suggestion
            has_production_name = production_names is not None and (existing_production_name := production_names.get(glyph.name))
            if has_production_name and "-" in existing_production_name:
```

.github/actions/import/scripts/fix_glyph_production_names.py
RickyDaMa: So production names only usually come from one UFO? Makes sense I guess, saves duplicating

madig:
Yeah, including it in the fixer seems appropriate, especially because it's supposed to be a bandaid for getting 1.0 out the door.

.github/actions/import/scripts/fix_glyph_production_names.py
RickyDaMa: Done. Ugly, but done

.github/actions/import/scripts/fix_glyph_production_names.py
RickyDaMa: Done

.github/actions/import/scripts/fix_glyph_production_names.py
```diff
@@ -26,15 +26,18 @@
...
             if not glyph.name:
                 continue
 
-            has_production_name = production_names is not None and glyph.name in production_names
-            if has_production_name and "-" in production_names[glyph.name]:
-                existing_production_name = production_names[glyph.name]
+            has_production_name = production_names is not None and (
+                existing_production_name := production_names.get(glyph.name)
+            )
+            if has_production_name and existing_production_name:
```
madig: Oh shi I dropped the `"-" in` 😬 sorry

RickyDaMa: I can't believe you've done this.

PR #218, opened 2023-01-09
Implement OT Sanitizer check

chrissimpkins:
Closes #217

Issue #217, opened 2023-01-09
Add OpenType Sanitizer check


chrissimpkins:
The Android team requested that we run OTS checks on the system fonts.

We have an implementation in GS @ https://github.com/googlefonts/googlesans/blob/e20bd20082e09ff2a700af5d60a9fb63afb5ffea/.github/workflows/cicd.yml#L111-L114

PR #219, opened 2023-01-10
Import workflow: add OTS check

RickyDaMa:
Follow up to #217 & #218, runs OT Sanitiser as part of the import workflow (otherwise wouldn't happen because of previously discussed workflows triggering workflows problems)

@chrissimpkins I presume this is something you'd like :) No need to merge otherwise

### DIFF
--- a/.github/workflows/import.yaml
+++ b/.github/workflows/import.yaml
@@ -84,6 +84,10 @@ jobs:
           cp -R "$GITHUB_WORKSPACE/main/qa" "$GITHUB_WORKSPACE/target"
           cd "$GITHUB_WORKSPACE/target" && make test
         continue-on-error: true
+      - name: Check with OT Sanitizer
+        uses: f-actions/opentype-sanitizer@v2
+        with:
+          path: "fonts/variable/*.ttf"
       - name: proof
         run: cd "$GITHUB_WORKSPACE/target" && make proof
       - name: Analyze file size


chrissimpkins:
The new workflow step [ran in the CI of the PR that I added yesterday](https://github.com/googlefonts/googlesans-flex/pull/218#issuecomment-1376437104).  I don't understand why it needs to be on the import workflow.  When a new imported commit lands in a new branch do we not trigger the build workflow?  Can you explain in a bit more detail?

RickyDaMa:
It depends on when you want OTS to run

If you use the import workflow and subsequently create a PR, OTS is not run on that PR without this PR being merged prior

Everything you want to be done on an import PR has to be included in the import workflow directly, as otherwise it's not automatically triggered. That's why the `.github/workflows/import.yaml` does it all: it imports, merges, commits, pushes, builds, tests, QAs, and runs file size analysis. It's not a good system, but it's what we're stuck with until we discuss solutions in a retrospective

chrissimpkins:
Got it!  Thank you very much Ricky.  Let's merge this.  Please let me know if you feel that it should be removed from the build.yml file.

RickyDaMa:
No harm in keeping it in the `build.yaml`. I'd rather keep that file maintained and then hopefully we can start using it properly again soon

PR #220, opened 2023-01-10
Build & import workflows: remove installation of yq, pin Python

RickyDaMa:
Unused as of #201 in build, never needed in import

Will save a bit of time & money

Did mean to do this in #201 but lose track of the removal in a rebase

Also pinned Python version in import, the same as for build

Safe and easy to merge @MariannaPaszkowska, although low priority


chrissimpkins:
Thanks Ricky!

PR #231, opened 2023-01-13
Import workflow: support importing commits

RickyDaMa:
@MariannaPaszkowska as requested in #117! Note that I would recommend you use the commit 594f926 as that's the one on fb-wip, whereas the one you linked me earlier was the merge commit from the import.

Technical jargon:

This actually simplifies the import workflow, just at the cost on giving up on shallow cloning the repo, since you can't `git fetch` a specific commit


MariannaPaszkowska:
Thank you @RickyDaMa!

PR #144, opened 2022-12-22
Update glyph list & check for missing or extraneous glyphs

RickyDaMa:
Willing to take suggestions on the name for this script, I can't think of a good concise one

Current output:
```
Missing the following glyphs (0):
-

The following glyphs in the source aren't in the import list (79):
- Eth-bar
- Euro.tf
- Oslash-bar
- a.alt
- approxequal.tf
- caroncomb.alt
- cent.tf
- colon.tf
- colon.time
- comma.tf
- currency.tf
- diagonalbarl
- divide.tf
- dollar.tf
- dot-composite
- eight.denominator
- eight.numerator
- eight.tf
- equal.tf
- exclam-bottom
- exclam-top
- five.denominator
- five.numerator
- five.tf
- flig1
- flig2
- four.denominator
- four.numerator
- four.tf
- g.alt
- greater.tf
- greaterequal-stroke
- greaterequal.tf
- horizontalbarlc
- less.tf
- lessequal.tf
- logicalnot.tf
- minus
- minus.tf
- multiply.tf
- nine.denominator
- nine.numerator
- nine.tf
- notequal
- notequal-slash
- notequal.tf
- numbersign.tf
- one.denominator
- one.numerator
- one.tf
- oslash-bar
- percent.tf
- percentbar
- percentbar.tf
- period.tf
- plus.tf
- plusminus.tf
- question-bottom
- question-top
- section.tf
- semicolon.tf
- seven.denominator
- seven.numerator
- seven.tf
- six.denominator
- six.numerator
- six.tf
- space.tf
- sterling.tf
- three.denominator
- three.numerator
- three.tf
- two.denominator
- two.numerator
- two.tf
- zero.denominator
- zero.numerator
- zero.tf
- zeropercent
```

PR #232, opened 2023-01-13
Import from `594f926`

RickyDaMa:

---

The workflow run (for fontbakery checks etc.) can be found [here](https://github.com/googlefonts/googlesans-flex/actions/runs/3910725641/jobs/6683262406)

The reason why there are so many deletions is because since this commit was last imported, the glyph list was updated (#144)

What you can [see in the logs](https://github.com/googlefonts/googlesans-flex/actions/runs/3910725641/jobs/6683262406#step:4:1475) is the postscript names being added as needed, as requested in #117


MariannaPaszkowska:
Closing because there are interpolation issues in this import:

Screenshot 2023-01-13 at 4 08 10 PM

Screenshot 2023-01-13 at 4 19 12 PM

@RickyDaMa @madig Let's work on the sources that are in the main; we need to update WinAscent, WinDescent, and update the glyph-names (addressing FontBakery fails from here https://github.com/googlefonts/googlesans-flex/actions/runs/3872071849/jobs/6600498713#step:11:232)

RickyDaMa:
@MariannaPaszkowska these issues should also appear on main too: there should be no change in the sources from what's on main, except the metadata fixes and removal of glyphs not on the import list. Can you confirm if these issues you've highlighted are on main?

dberlow:
Because “not =“ is a composite of = and a bar, I see no possibility that this came out of our sources, unless people are decomposing composites and merging them from different designspace locations.On Jan 13, 2023, at 9:23 AM, Marianna Paszkowska ***@***.***> wrote:﻿
Closing because there are interpolation issues in this import:


@RickyDaMa @madig Let's work on the sources that are in the main; we need to update WinAscent, WinDescent, and update the glyph-names (addressing FontBakery fails from here https://github.com/googlefonts/googlesans-flex/actions/runs/3872071849/jobs/6600498713#step:11:232)

RickyDaMa:
> we need to update WinAscent, WinDescent, and update the glyph-names

This should all be done now thanks to Chris in #237 

> I see no possibility that this came out of our sources, ...

Apologies if my statement came across as accusatory! My point was that since we had already imported the sources and they were in main, if we were importing the same sources again (which is what this PR does) *if our process worked as expected* we would not expect the unchanged sources to change. Therefore it was my expectation that the issue should be present on `main` as well as they were not intended changes in the import workflow.

> Can you confirm if these issues you've highlighted are on main?

I just wanted this clarified as then it would help us start investigating whether the issue was in our import process or not. If there are discrepancies between untouched glyphs (like not equals) in main and this PR, that indicates a regression or inconsistency in the import script that I'd want to investigate

PR #237, opened 2023-01-13
Run .github/actions/import/scripts/fix_metadata_in_sources.py on regular directory source files

chrissimpkins:


chrissimpkins:
@MariannaPaszkowska @RickyDaMa 

postscript name and v metrics changes as expected with the script run?

RickyDaMa:
Beat me to it! I'm still interested to know if we can't use #232, but this PR is less diff-heavy and easier to verify. Postscript changes look fine, @madig are the vertical metrics as you expect? I don't see anything in fontbakery that complains about them, so that's a good sign

chrissimpkins:
> Beat me to it! I'm still interested to know if we can't use #232, but this PR is less diff-heavy and easier to verify...

We'd like to avoid any further glif design updates in the RC builds.  There has been a great deal of work going on in the fb-wip branch this week and we won't have time to review it against the review that Marianna performed earlier this week to approve for the first Android push.

MariannaPaszkowska:
Thank you @chrissimpkins @RickyDaMa! WE can look into what went wrong [here](https://github.com/googlefonts/googlesans-flex/pull/232) next week

PR #247, opened 2023-01-17
Specify font version in config.yaml

RickyDaMa:
Low priority PR, but now that we're using the latest & greatest gftools, we might as well take advantage of being able to specify the font version directly in the config.

~~I've also updated the metadata fixing script to read this version value instead of being hardcoded~~

(Addendum: internal discussion about storing version as string or float [link](https://daltonmaag.slack.com/archives/G010DM236JJ/p1673951030651279))

### DIFF
--- a/sources/config.yaml
+++ b/sources/config.yaml
@@ -7,7 +7,7 @@ axisOrder:
   - wght
   - ROND
 familyName: Google Sans Flex
-# TODO: specify version once the option is available
+version: "1.002"
 # The existing UFO sources are quadratics, so it probably does not make sense
 # to build OTFs. Also, Google Sans doesn't build OTFs, so it makes sense to do
 # the same for Flex


MariannaPaszkowska:
@RickyDaMa are there any updated on this?

RickyDaMa:
Hey @MariannaPaszkowska, I've updated the PR for you

We're using a supported version of `gftools` now that has the feature. It appears to work as part of the build, however I noticed that the unique font identifier was still 1.000... (see image). Didn't know if that was a concern - not sure where that's set/generated currently. Fontbakery doesn't take issue with it

image

[Fonts.zip](https://github.com/googlefonts/googlesans-flex/files/11756995/fonts.zip)

MariannaPaszkowska:
@RickyDaMa In the next release, the unique font identifier should be 1.002. Could we update this PR?

RickyDaMa:
Hey @MariannaPaszkowska, I've done some experimenting with the behaviour of this PR today. Currently, the CI workflow during a release sets the version using `bumpfontversion`, which handles all the fields in my previous screenshot without issue. Setting `version` in the `config.yaml` (i.e. this PR) overwrites this, but not for the unique font identifier, meaning that you can get mismatching/conflicting fields. Thus, I'm going to close this PR for now, as the current system works, and this PR could break it.

I guess it's a shortcoming of `gftools` that it doesn't modify the unique font identifier with the new version. I could open an issue on the repository about this if you'd like to move this forward and revive this PR one day.

PR #249, opened 2023-01-17
Add commit ID to font

### LINKED ISSUE #245: Add build git commit SHA1 short hash to the name ID 5 record
Let's add the build git commit short SHA1 hash to the nameID 5 record following a semicolon.

madig:
Closes https://github.com/googlefonts/googlesans-flex/issues/245.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,8 @@ venv: venv/touchfile
 build.stamp: venv sources/config.yaml $(SOURCES)
 	rm -rf fonts/
 	venv/bin/gftools builder sources/config.yaml
-	venv/bin/font-v write --sha1 $(shell find fonts/variable -type f)
-	venv/bin/python scripts/sanitize-name-table.py $(shell find fonts/variable -type f)
+	venv/bin/font-v write --sha1 fonts/variable/*.ttf
+	venv/bin/python scripts/sanitize-name-table.py fonts/variable/*.ttf
 	touch build.stamp
 
 venv/touchfile: requirements.txt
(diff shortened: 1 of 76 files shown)


Makefile
```diff
@@ -19,7 +19,10 @@ build: build.stamp
 venv: venv/touchfile
 
 build.stamp: venv sources/config.yaml $(SOURCES)
-	. venv/bin/activate; rm -rf fonts/; gftools builder sources/config.yaml && touch build.stamp
+	rm -rf fonts/
+	venv/bin/gftools builder sources/config.yaml
+	touch build.stamp
+	venv/bin/font-v write --sha1 $(shell find fonts/variable -type f)
```
RickyDaMa: `touch build.stamp` should be the last thing done here I believe, because that's what Make uses to detect if the step has been completed or needs redoing.

As it currently stands, this *could* lead to the situation where `font-v` is interrupted or isn't run but Make thinks everything is already built when it's next run

RickyDaMa (commented):
General question: why don't you activate the venv anymore during the build step? I presume gftools handles this itself, but still

madig:
I don't think it does. You can activate a venv or just run a bin inside it directly; you need to activate for every new line then anyway. Needed to put these onto their own lines because the shell command came back empty-handed for some reason (race condition?). 🤷‍♀️

RickyDaMa (approved):

madig:
@MariannaPaszkowska I added a build step to replace non-letters and non-numbers in the version name record with underscores.

MariannaPaszkowska:
@madig What's the blocklist of those characters?

madig:
Everything that is not in `abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789.`.

MariannaPaszkowska:
Thanks, please add also `,` `;` and `space` to the allowlist.

Issue #238, opened 2023-01-13
Add ROND axis width checks


chrissimpkins:
Please also add width checks to *-ROND0 to matching *-ROND100, as I have requested that check to be run on our commits as well.

E.g. GoogleSansFlex-opsz18-wdth100-wght400-GRAD0-ROND0 must match GoogleSansFlex-opsz18-wdth100-wght400-GRAD0-ROND100

_Originally posted by @dberlow in https://github.com/googlefonts/googlesans-flex/issues/182#issuecomment-1381808783_

PR #251, opened 2023-01-17
Add ROND reflow check

madig:
Closes https://github.com/googlefonts/googlesans-flex/issues/238.

Copy-pasted and adapted from the upstream GRAD check.


MariannaPaszkowska:
Brilliant, thank you @madig!

PR #257, opened 2023-01-19
Fail when fontbakery fails

### LINKED ISSUE #254: Use fontbakery exit status codes in the GitHub Actions CI tests
The non-zero fontbakery exit status code [is currently defined as acceptable](https://github.com/googlefonts/googlesans-flex/actions/runs/3953017972) and CI passes despite failing tests to allow the completion of the workflows.  Let's confirm that we toggle the non-zero exit status code back on so that the CI fails in advance of our March release when we are comfortable that we have either addressed the failing issues or ignore the fontbakery check as not relevant in this project.

madig:
Closes https://github.com/googlefonts/googlesans-flex/issues/254.

This also means that artifacts aren't uploaded when fontbakery fails.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -41,7 +41,6 @@ jobs:
         run: make build
       - name: Check with fontbakery
         run: make test
-        continue-on-error: true
       - name: Check with OT Sanitizer
         uses: f-actions/opentype-sanitizer@v2
         with:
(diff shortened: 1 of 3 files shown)


madig:
@MariannaPaszkowska this would be "ready" as soon as the sources are fixed.

MariannaPaszkowska:
@RickyDaMa apologies, can we rebase this again?

PR #266, opened 2023-01-24
Add import from layers

### LINKED ISSUE #264: Build fails - Sparse layers not yet supported
https://github.com/googlefonts/googlesans-flex/actions/runs/3985057886/jobs/6831976632

madig:
This adds importing from layers to the import process, making sparse masters/sources/layers possible.

A side-effect is that the name of the default layer is imported as well, which has no functional consequence.

Closes https://github.com/googlefonts/googlesans-flex/issues/264.


madig:
@chrissimpkins @MariannaPaszkowska ready for review.

MariannaPaszkowska:
Thank you @madig!

PR #273, opened 2023-01-27
Add more features

madig:
* Hook up ss01 and ss02 features for `a` and `g`.
* Remove spurious unused instance

@MariannaPaszkowska


chrissimpkins:
Ty! 🚀 

6JzL79pjATSkvtz

madig:
Mind you that features currently aren't imported automatically, so if FB writes something, we need to import it manually.

madig:
@MariannaPaszkowska @chrissimpkins I think this is ready for merging as of right now; the upstream sources have more of the needed glyphs, but as of right now, they don't build. The features can be enabled as soon as the sources are importable again.

chrissimpkins:
~ superceded by #288

PR #279, opened 2023-01-30
Build at latest commit with compatible outlines

chrissimpkins:
Imported with import workflow at commit https://github.com/googlefonts/googlesans-flex/commit/e24aada3ee71fd82d9bdc7c5550451df5f486ae9 before incompatible glyphs introduced in the git history


chrissimpkins:
File for review is available at https://github.com/googlefonts/googlesans-flex/suites/10662882018/artifacts/533054374

madig:
Is this now outdated?

PR #282, opened 2023-01-31
Add collidoscope

madig:
Add collidoscope testing as separate command: Run `make run-collidoscope` when you need it and have a look at the HTML report in `out/fontbakery`.

The dependencies needed to be updated because collidoscope was outdated and not finding anything. The tool has an `area` parameter that can be used to tweak the threshold of how much overlap is needed to be considered an actual overlap. It might need some fiddling.


madig:
I think fixing Collidoscope to make area theshold checking possible would take longer, see https://github.com/googlefonts/collidoscope/issues/14.

madig:
@chrissimpkins @MariannaPaszkowska Pulling this in requires a dependency update. Is GSF ready for that? Also, the update of ufo2ft means it will crash on current main sources because `foursuperscript` and `foursuperior` have the same Unicode assigned, which was apparently fixed in the italics branch.

PR #286, opened 2023-02-01
Hook up diffenator

madig:


.github/workflows/diffenator.yaml
```diff
@@ -0,0 +1,28 @@
...
+          restore-keys: |
+            ${{ runner.os }}-venv-
+      - name: Build font
+        run: make build
+      - uses: f-actions/diffenator2@main
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-before: "github-release"
+          run-diffenator: true
+          path: "fonts/variable"
```
m4rc1e:
```suggestion
      - uses: f-actions/diffenator2@main
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
          fetch-before: "github-release"
          run-diffenator: true
          path: "fonts/variable/*.ttf"
          path-before: "fonts/variable/*.ttf"
```

m4rc1e:
I don't mind pushing this across the line if it fails again. I'd just look at the diffenate job in Google Sans, https://github.com/googlefonts/googlesans/blob/main/.github/workflows/diffenate.yml#L78-L85

requirements.txt
```diff
@@ -170,7 +170,7 @@ gitdb==4.0.9
     # via gitpython
 gitpython==3.1.29
     # via font-v
-glyphsets==0.5.4
+glyphsets>=0.5.4
```
madig: Please don't. Use `make update` to pin new deps, maybe basing your PR on https://github.com/googlefonts/googlesans-flex/pull/500.

m4rc1e: @madig I'm just hacking so I actually get the results.

m4rc1e: Will clean up once we get a downloadable archive with all diffs.

m4rc1e: Boom. I've finally got an archive of tests, https://github.com/googlefonts/googlesans-flex/actions/runs/4617508954. This was a pita to figure out so it's pretty clear that I need better exceptions and logging.

@madig @MariannaPaszkowska are you interested in browser diffs as well? personally, I would recommend em.

m4rc1e: Couldn't help myself. You'll now get images for the Thin, Regular and Black. Lmk if you want more instances.

diffbrowsers_text-Windows-10-10 0 20348-SP0_chrome_111 0 5563 111

m4rc1e: @madig I've just rebased and included #500.

m4rc1e:
@madig this is ready now. I just spoke to @chrissimpkins and @MariannaPaszkowska and for the time being we agreed that this workflow should only run when we push branches prefixed with "diff-". This matches the Google Sans implementation

PR #306, opened 2023-02-03
Add OpenType sups feature support

chrissimpkins:


chrissimpkins:
sups functions appropriately

PR #307, opened 2023-02-03
Add OpenType subs feature support

chrissimpkins:


chrissimpkins:
subs functions appropriately

PR #308, opened 2023-02-03
Add OpenType zero feature support

chrissimpkins:


chrissimpkins:
The substitution functions appropriately but there is a bug in the outline @ width values > 100 (reported as issue #309)

PR #310, opened 2023-02-03
Add OpenType ss01 and ss02 feature support for alternates with diacritics

chrissimpkins:
Expands the ss01 supported alternate-a repertoire and ss02 supported alternate-g repertoire

Closes #311

PR #312, opened 2023-02-03
Update glyph-list - Remove gcircumflex.alt

MariannaPaszkowska:
Removing gcircumflex.alt, because the gcircumflex glyph is not present in the sources and was not a part of the defined character set for this version.


chrissimpkins:
I confirmed that this does not show up in the feature code that we are working on in #310.

chrissimpkins (approved):

PR #329, opened 2023-02-06
Point UFOs to shared family.fea

madig:
This makes UFOs always point to the family.fea file next to the target Designspace to import into. All feature development should happen inside family.fea, and all changes need to be ported by hand.

### DIFF
--- a/.github/actions/import/scripts/gs_merge_designspace.py
+++ b/.github/actions/import/scripts/gs_merge_designspace.py
@@ -33,6 +33,7 @@ import copy
 import logging
 import shutil
 import sys
+import os.path
 from pathlib import Path
 from typing import Dict, List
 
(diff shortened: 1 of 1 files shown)


chrissimpkins:
Thanks Nikolaus!  Let's merge and try it.

chrissimpkins:
I ran the updated importer against the latest fb-wip branch in https://github.com/googlefonts/googlesans-flex/actions/runs/4105413587

chrissimpkins:
Mind deleting your branch here if it works as intended?

madig:
Awww, incompatibilities again... Well, it works when testing locally, on fb-wip from 2 commits ago.

madig:
Oh, just to make it clear, this will not import features, just point all UFOs to a shared family.fea file. Features need to be written and imported manually, there's no good way to automate this.

chrissimpkins:
> Oh, just to make it clear, this will not import features, just point all UFOs to a shared family.fea file. Features need to be written and imported manually, there's no good way to automate this.

Np at all.  All features are implemented in the top level family.fea file and we'll manually edit the file with changes.  The goal was only to make sure that the feature code was imported into the masters via the individual UFO master features.fea files as required to compile them.

PR #320, opened 2023-02-03
Implement optical size axis avar1 map (Black Foundry approach)

chrissimpkins:
For comparison with #335 

This PR adds the Black Foundry optical size axis avar1 map that is defined as follows:

```xml
<map input="6" output="0"/>
<map input="8" output="0.004"/>
<map input="14" output="10"/>
<map input="18" output="35.2"/>
<map input="24" output="73"/>
<map input="92" output="93"/>
<map input="144" output="100"/>
```

avar table data:

Stepwise non-linear interpolation (user values on the x-axis):

chart (1)

This is meant to support builds for review of this issue in the Latin glyph set and is not necessarily the decided approach to opsz avar.  Opened as a draft PR while we discuss this.

PR #335, opened 2023-02-06
Implement optical size axis avar1 map (Font Bureau approach)

chrissimpkins:
For comparison with #320

The avar1 map is defined as follows:

```
<map input="6" output="6"/>
<map input="18" output="18"/>
<map input="92" output="120"/>
<map input="144" output="144"/>
```

avar table definition:

Stepwise non-linear interpolation (user values on the x-axis):

chart (2)

cc @dberlow @sannorozco @MariannaPaszkowska

### DIFF
--- a/sources/regular/GoogleSansFlex.designspace
+++ b/sources/regular/GoogleSansFlex.designspace
@@ -8,7 +8,11 @@
       <labelname xml:lang="en">width</labelname>
     </axis>
     <axis tag="opsz" name="opsz" minimum="6" maximum="144" default="18">
-      <labelname xml:lang="en">optical size</labelname>
+      <map input="6" output="6"/>
+      <map input="18" output="18"/>
+      <map input="92" output="120"/>
+      <map input="144" output="144"/>
+      <labelname xml:lang="en">Optical size</labelname>
     </axis>
     <axis tag="ROND" name="ROND" minimum="0" maximum="100" default="0">
       <labelname xml:lang="en">ROUNDED</labelname>


chrissimpkins:
@dberlow @sannorozco @MariannaPaszkowska A test font with this avar implementation is available in this zip archive: https://github.com/googlefonts/googlesans-flex/suites/10812435141/artifacts/544011484

chrissimpkins:
@madig Any idea how this avar definition landed in the main branch?  We appear to now define an avar table on the opsz axis as of https://github.com/googlefonts/googlesans-flex/commit/4a503c73bb12241ef29081b45c8aa05631cf6b01

https://github.com/googlefonts/googlesans-flex/blob/232b3381106d71e045d4c9961b662fa55901d771/sources/regular/GoogleSansFlex.designspace

madig:
It's coming from a recent import: https://github.com/googlefonts/googlesans-flex/blame/232b3381106d71e045d4c9961b662fa55901d771/sources/regular/GoogleSansFlex.designspace#L4-L10

Since sources are still unstable, we always replace the Designspace with whatever the import has.

chrissimpkins:
@dberlow @sannorozco Will you please remove the opsz avar1 definition from your fb-wip branch until we agree on the definition?  We're automatically importing it into sources that we are using for review.  Ty!

dberlow:
Spoke with Marianna yesterday about finding a difference by looking around. Here are sizes from 
On the left is with avar changes, right is without. 
Screen Shot 2023-02-15 at 5 39 27 AM

There is a slight difference in weight you caan see, with the avar-ed version on the left. If the avar attempting to bring the effects of opsz144 down further in the opsz range, the left should be lighter than the right, but it is not. 

And here is 44-65 and on the left is with avar changes, right is without. 
Screen Shot 2023-02-15 at 5 02 17 AM

It too is doing the opposite of what I expect. 

(I think we should go without it).

dberlow:
Summary: the current GSF Latin includes weight width and optical size axes. It has been attempted to apply an avar, copied to the current builds of GSF from the avar for CJK. Very little difference was notable between the product with and without the avar changes of the CJK version.

I have looked at the results and advise against such an attempt at this time. I think it is better not to make changes to the current GSF unless there is a design reason for it, except for a technical requirement. During and after the current font becomes a test-product progressing though to acceptance, we can reevaluate based on change requests in that process.

Background: There are a lot of issues surrounding the idealization of the luminance and alignments among scripts. GSF for Latin has been designed with extremes, (opsz6-144,wght1-1000 and wdth25-151), making it likely most all other GSF scripts will fit that designspace or be a subset of it.

To the GSF Latin design space, we have added a grade axis and are scheduled to add a set of parametric axes giving alignment and width refinement to the Latin. These and grade are intended for multiple uses, one of them being to help match the Latin to itself when a production process interferes with wght, via grades. Along with this, the ability of Latin to adjust to the alignments of other scripts, and to make possible the adjustment of the Latin to the weight and wdth ranges of other scripts will be enabled.

So the global GSF strategy is to make the defaults of all the scripts match Latin in weight, (at the default for all scripts, opsz18,wdth100,wght400), with whatever their natural alignments for reading that script are, at that size, weight and width. Then for each other script, tactics include grades and axes for alignments and widths required to be compatible with Latin. 

That future More flexible GSF Latin, also enables disparities between the design spaces to be adjudicated in general, or in script-r language-specific versions that may be desired. This leaves a GSF font project with the ability to go in any direction to provide multi-lingual support of compatible text, customizable at several levels and over time.

Concluding: We have a Latin that’s good and broad for Latin styling and use in design and are working toward a future of more flexibility. Making changes to the avar now seems like taking a change to the optical sizes via avar at this time could be causing more disturbance today and tomorrow, than solving any issue in particular today

PR #340, opened 2023-02-07
Add check for tabular glyph widths

madig:
Add a source check for ensuring that tabular glyphs (`.tf`) all have the same width.


MariannaPaszkowska:
@chrissimpkins, we added this check to confirm that all tabular glyphs have the same width within the same master.

chrissimpkins:
Great job!  Good check

PR #408, opened 2023-02-21
Add overlap bit setter

### LINKED ISSUE #382: Set glyf table OVERLAP_SIMPLE and OVERLAP_COMPOUND bit flags
Teams are seeing overlaps in Android environment-specific stroked outlines with this family.  We introduced the overlap bit flag change to address it in Google Sans around v8.001 and I think that we may be able to address it here with the same script if we are not currently doing so?

Google Sans Issue: https://github.com/googlefonts/googlesans/issues/478
@madig PR to add script in the GS build workflow: https://github.com/googlefonts/googlesans/pull/494

Let's plan to include this in the production build workflow for the v1.1 release.

madig:
Copied from GS.

Closes https://github.com/googlefonts/googlesans-flex/issues/382.


chrissimpkins:
@madig do we know if this is appropriately labeling outlines with overlaps?

madig:
It did for GS, I have not done an in-depth check for GSF.

PR #409, opened 2023-02-21
Add kerning pair counter check

### LINKED ISSUE #407: Check: report number of kerning pairs in all masters
We want to know if there are any masters that are missing kerning, and report the number of pairs.

madig:
Closes https://github.com/googlefonts/googlesans-flex/issues/407

@MariannaPaszkowska

### DIFF
--- a/qa/check-sources.py
+++ b/qa/check-sources.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 from typing import Iterable, Tuple, Union
 
 from fontbakery.callable import check, condition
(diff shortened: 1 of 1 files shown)


MariannaPaszkowska:
@madig An afterthought: In this kerning check, should we maybe skip the intermediate masters, ex: GoogleSansFlex-opsz17.999-wdth25-wght400-ROND0.ufo . I think they should not have kerning. What do you think?

madig:
Done in #410

PR #410, opened 2023-02-21
Skip sparse sources when counting kerning pairs

madig:
@MariannaPaszkowska


MariannaPaszkowska:
@madig Would it be easy also to skip the Skateboard Previews folder?

MariannaPaszkowska:
Fantastic, thank you @madig

PR #421, opened 2023-02-24
Set overlap bits via list

madig:


MariannaPaszkowska:
Related issue: https://github.com/googlefonts/googlesans-flex/issues/382

PR #500, opened 2023-04-03
Update dependencies

madig:
Trying to fix the CI.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,6 +8,7 @@ on:
 jobs:
   build:
     runs-on: [linux, googlefonts-64cores-256GB]
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python 3.11
--- a/.github/workflows/import.yaml
+++ b/.github/workflows/import.yaml
@@ -13,6 +13,7 @@ on:
 jobs:
   import-branch:
     runs-on: [linux, googlefonts-64cores-256GB]
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v3
         with:
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ build.stamp: venv sources/config.yaml $(SOURCES)
 venv/touchfile: requirements.txt
 	test -d venv || python3 -m venv venv
 	. venv/bin/activate; pip install -U setuptools wheel pip
-	. venv/bin/activate; pip install -r requirements.txt
+	. venv/bin/activate; pip install --no-deps -r requirements.txt
 	touch venv/touchfile
 
 test: venv build.stamp
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ update-shaping-expectations:
 
 update:
 	pip install -U pip-tools
-	pip-compile -U requirements.in
+	pip-compile --resolver=backtracking -U requirements.in
 
 file-size: build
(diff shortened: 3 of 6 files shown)


madig:
@MariannaPaszkowska @chrissimpkins :)?

madig:
My best bet on why it blew up before is that pip-compile does not pip git-dependencies to a commit, see https://github.com/googlefonts/googlesans-flex/pull/500/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552L103. We pulled in gftools at a certain commit, which in turn pulled in a git-dependency on diffenator2 without pinning. Since that's a moving target, it led to a dependency conflict with the other pinned deps.

After consulting https://pip.pypa.io/en/stable/topics/dependency-resolution/#possible-ways-to-reduce-backtracking, I decided to make pip not resolve dependencies during installation (`--no-deps`), because pip-tools does it for us when we update dependencies. This should fail fast if something is wrong. I'll also remove the time limit again.

chrissimpkins:
Thanks for having a look at this Nikolaus.  I really appreciate it.  

> I decided to make pip not resolve dependencies during installation (--no-deps), because pip-tools does it for us when we update dependencies.

Let's document our approach to dependency definitions.  It sounds like this makes us dependent on use of pip-tools.  That may be ok.  But let's make sure that anyone who travels across the need to build this project is aware that you don't manually edit the requirements file because we intentionally do not resolve dependency versions with pip.

> I'll also remove the time limit again.

How did you define a time limit? If this is possible, I recommend that we keep it and set it to ~ 60 mins.  That should be more than enough for builds in this project unless we begin to merge other scripts at which point we can reassess.

chrissimpkins:
It looks like these updates touch on compiler library dependencies.  I suggest that we have a look at the binaries to understand how they change across the compiler versions before we merge it.

chrissimpkins (changes requested):
Will you please have a look at the binary diff that takes place with the compiler dependency updates?  We now have a production version pushed and will need to understand regressions against prod.

madig:
I see 

1. a name table diff because the version string has a different commit ID
2. Inscrutable GDEF, GPOS and GSUB diffs that I suppose come from
    1. The new kerning splitter
    2. https://github.com/fonttools/fonttools/pull/2974 merging duplicate lookups (remember https://github.com/googlefonts/googlesans/issues/567?)

So, maybe https://github.com/googlefonts/googlesans-flex/pull/286 brings more clarity here. @m4rc1e? :)

madig:
I used https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes to set a timeout.

m4rc1e:
Hey folks,

In #286, I've included this PR which created the following diffs, https://github.com/googlefonts/googlesans-flex/actions/runs/4619101201.

I can see the following changes:
- GRAD axis added
- no wordlist regressions for any fvar instances
- GPOS, GDEF, MVAR have changed as Niko pointed out. However, the changes haven't caused any word regressions so I'm happy to pass these

I think for my sanity, I'm going to implement the ability to change the diffenator2 diff threshold today, https://github.com/googlefonts/diffenator2/issues/37.

So far LGTM but let me check first with a lower diffenator2 tolerance.

m4rc1e:
I manually changed the tolerance down from 0.9 down to 0.0001 and there are no wordlist differences. LGTM

I'll add the ability to change the threshold to diffenator2 next week.

PR #506, opened 2023-05-16
Add ufo2glyphs2ufo scripts

madig:
Copy the scripts from GS.

I converted the uprights to .glyphs and loaded them into Glyphs 2.6.x without problems. I had to save the file there before I could open it in Glyphs 3, because the ascender fontinfo value for one of the sources is a floating point value when it shouldn't be, but that can also be fixed by hand.

I suppose we can go with the GS merge workflow, which I can work on when we get to it.


madig:
@MariannaPaszkowska I added a blurb to the Readme.

PR #544, opened 2023-06-07
Georgs test branch

schriftgestalt:
This is not supposed to be actually merged!! Only to discuss those changes.


MariannaPaszkowska:
@madig - Georg mentioned to me that you already reviewed those UFO files saved from glyphs and that the differences are not significant; anything you's like to add? 
CC @chrissimpkins

MariannaPaszkowska:
@chrissimpkins, Georg reported that the UFO quadratic support in Glyphs is ready (for single files). I reviewed them with Georg yesterday, and I think we are ready to move the production to Glyphs. In this PR, there are the UFO files from our repo that have been saved in Glyphs. 
CC @octaviopardo

PR #545, opened 2023-06-08
Import from it-ad-wip-italic-v1.002

RickyDaMa:
Creating this PR to demonstrate how the import workflow currently handles the italics branch (`it-ad-wip-italic-v1.002`) so we can see if any adjustments are needed. The good news is that it ran without errors etc., so that's positive!

Workflow run can be found [here](https://github.com/googlefonts/googlesans-flex/actions/runs/5209545511/jobs/9399519365)

### DIFF
--- a/sources/italic/GoogleSansFlex-Italic.designspace
+++ b/sources/italic/GoogleSansFlex-Italic.designspace
@@ -1037,7 +1037,7 @@
     </source>
   </sources>
   <instances>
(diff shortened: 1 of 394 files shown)


madig:
@MariannaPaszkowska note that we import some regular sources here. Nothing actually important from a spot check, but you'd need to double-check each import PR to see if the team changed something they shouldn't have.

PR #547, opened 2023-06-09
Add print space width

madig:
Add one-off script to print the advance width of the space glyph in a VF.


MariannaPaszkowska:
@madig Can I ask you to add the Apache license header?

PR #548, opened 2023-06-12
Update burndown script for italics

madig:
The italics branch seems to have usable color marks, so here's the updated script to track italic progress.

The script currently assumes we have 2 * 120 UFOs, is there a newer estimate? There are 2 * 128 in total, of which 162 are full UFOs, and the rest are sparse. @MariannaPaszkowska 

gs-flex-progress


RickyDaMa:
@MariannaPaszkowska do you want a GitHub workflow (a manual one) for the burndown chart, or are you happy to run it locally? It takes quite a long time to run so could end up being costly. The above example chart only analyses 18 commits on Nikolaus' local machine and still takes 4 minutes to run. As the italic branch keeps getting more development this will only increase

RickyDaMa:
Fixed the trend line name, and the legend labels

gs-flex-progress

RickyDaMa:
Now with more granular milestones

gs-flex-progress

Doing something pretty like having the Alpha line start where the Prototype line ends is definitely possible but would require more work. This is the quick 30 minute version 😄

MariannaPaszkowska:
It looks much better. Great job! Implementing that pretty target line, where one milestone's line starts and the other ends, sounds great! And possibly in a flexible way that's easy to adjust (so we can reuse this tool easily). Can you please send me an effort estimate?

RickyDaMa:
Here's the fancy line!

gs-flex-progress

As for the bigger "let's make this a separate open source tool that's user configurable" - we're currently discussing internally what that looks like and how we might go about it. Stay tuned 😉

MariannaPaszkowska:
Exciting! Thanks a lot!

MariannaPaszkowska:
@RickyDaMa Can we add the first milestone too (concept -> June 1st, 13 glyphs:
HAKOGgnoakv /period /space)?

RickyDaMa:
Done!

gs-flex-progress

RickyDaMa:
Hi @MariannaPaszkowska, would it be possible to get clarification on the number of UFOs we're expecting the project to have? It'd make the burndown charts a lot more accurate, as currently we're guessing the number, meaning we may be under/over estimating, thus making the graph misleading.

GS Flex currently has 162 non-sparse masters. The script assumes there will be 240 masters at each milestone. Would it be possible for you to clarify what the expected number of masters would be at each milestone please?

MariannaPaszkowska:
How about we include 162 non-sparse masters for all glyphs and the sparse masters for the related glyphs only? 
@octaviopardo

octaviopardo:
Not sure if this is what you ask but here it comes:
The masters in the italic (and I assume in the upright as well, so it is a matter of multiplying per 2) are
ROND100: 39 masters, all of them Key
Sparse masters: 46
Key masters: 42,
TOTAL: 127
so in the full font it should be 254, but the 127 from the upright will not be colored, so you should not account for those.

RickyDaMa:
At 162 masters, this is what the burndown chart looks like

Sparse masters & glyphs shouldn't make too much of a difference since they're so small in proportion to the number of glyphs in a full master * the number of masters

gs-flex-progress

RickyDaMa:
Looking just at italics masters with 127 masters at all milestones (sorry Octavio your message didn't load for me until just now):

gs-flex-progress

octaviopardo:
Nice! This makes much more sense. Congrats on the setup @RickyDaMa Very nice control tool

MariannaPaszkowska (approved):

MariannaPaszkowska:
Thank you, everyone! Fantastic work 🥳

PR #570, opened 2023-06-20
Work on metadata of VFs

madig:
This changes the import scripts to fix up the metadata on the italic sources better when importing a fresh batch.

You should be able to install upright and italic font at the same time, but not the "Full" font, which I suppose is temporary?

@MariannaPaszkowska


MariannaPaszkowska:
Thanks a lot! Yes we want to install either Italic and Roman or one font that combines Italic and Roman. 
Cc @chrissimpkins

chrissimpkins:
> Thanks a lot! Yes we want to install either Italic and Roman or one font that combines Italic and Roman. 
> Cc @chrissimpkins 

The combined font will support Italic on a slant axis

madig:
Note that we use ital in the single fonts but slnt in the combined one. We should use slnt in both.

PR #695, opened 2023-08-15
Move all ROND=100 masters to root of the regular and italic source directories

### LINKED ISSUE #694: [fontc] use a single fea file
```shell
$ rm -rf build/ && time target/release/fontc --source ../googlesans-flex/sources/GoogleSansFlex.designspace 
[2023-08-14T22:47:40.326681Z main fontc::workload ERROR] Fe(Features) failed Expected to match: "../googlesans-flex/sources/regular/GoogleSansFlex-opsz6-wdth25-wght1-ROND0.ufo/features.fea", "../googlesans-flex/sources/regular/ROND100/GoogleSansFlex-opsz6-wdth25-wght1-ROND100.ufo/features.fea"
Error: TasksFailed([(Fe(Features), "Expected to match: \"../googlesans-flex/sources/regular/GoogleSansFlex-opsz6-wdth25-wght1-ROND0.ufo/features.fea\", \"../googlesans-flex/sources/regular/ROND100/GoogleSansFlex-opsz6-wdth25-wght1-ROND100.ufo/features.fea\"")])
```

This occurs because GSF doesn't point to the same fea file everywhere. They are equivalent but you'd have to parse them to figure it out. Per IM with Chris lets make it use the same one.

https://github.com/googlefonts/googlesans-flex/compare/main...make-fontc-build has a fix for this, combined with some other change I'm less familiar with.

chrissimpkins:
Closes #694 

cc @rsheeter @anthrotype


chrissimpkins:
@MariannaPaszkowska @octaviopardo @RickyDaMa 

Any concerns about moving all of the `regular/ROND100` and `italic/ROND100` masters to the root of the respective source directories as we've done here?  The current version of the fontc compiler does not traverse up directory trees to locate feature files so we can't drop the UFOs into sub-directories below the level of the root feature file import path.  IIUC this is something that Ricky was working on in https://github.com/googlefonts/googlesans-flex/commit/2c32c0a2fe90139ad88ec27aded19ebb13a73e7a

octaviopardo:
Nope on my side

chrissimpkins:
@RickyDaMa What do we need to do to trigger the build and CI checks on this PR branch?

RickyDaMa:
To clarify, the issue isn't that `fontc` can't traverse to these directories, it's moreso that it can't assert that the feature files are the same, because the file contents is different, even though the feature file (after resolving includes) would be identical

See my comment [here](https://github.com/googlefonts/fontc/issues/72#issuecomment-1663774594) on the `fontc` repo about this

@chrissimpkins all builds are now triggered manually [here](https://github.com/googlefonts/googlesans-flex/actions/workflows/build.yaml), I've set a run off for this PR [here](https://github.com/googlefonts/googlesans-flex/actions/runs/5877094505)

anthrotype:
yeah, we just compare the textual content is the same, to be able to know that the includes resolve to the same file we'd have to parse them, we prefer not to do that at this stage

RickyDaMa:
The build isn't happy by the looks of it; I'll look into this today

RickyDaMa:
See #697 as a replacement for this PR. My `make-fontc-build` branch had more already-done work on it for this migration (this PR does not update feature files to adjust the relative paths), plus I have also now fixed the smaller designspaces (also not present here). Closing this PR to help avoid confusion :)

chrissimpkins:
Thanks Ricky!

PR #700, opened 2023-09-12
Drop name table sanitizer script

madig:
I don't remember why I added this, but it only causes

```
    <namerecord nameID="5" platformID="3" platEncID="1" langID="0x409">
      Version 1.002;[fa35e8d625];gftools[0.9.27]
    </namerecord>
```

to turn into

```
    <namerecord nameID="5" platformID="3" platEncID="1" langID="0x409">
      Version 1.002;_fa35e8d625_;gftools_0.9.27_
    </namerecord>
```

and `fontbakery check-googlefonts` does not complain about the former.


chrissimpkins:
This addressed a bug in the Font Bureau video proofer application that is now fix.  We should not need the workaround any longer.  Good to remove from my standpoint.

cc @rsheeter this removes the `sanitize-name-table.py` script that we discussed yesterday

PR #718, opened 2023-11-06
WIP: [do not merge] Repopulate sources/ from V2 import of Glyphs 3 sources

Hoolean:
This PR contains a work-in-progress import of the V2 sources from the Glyphs 3 files in `it-ad-wip-v2.0`.

We do not need to review or merge this branch yet, but it should let us produce a build on the CI and plan our next steps, especially with regards to automating future imports 🚀 

Although GitHub is bravely trying, the diff is not easily viewable, however, the change is split into two parts (see commit descriptions for greater detail):

1. 5762c2e712de4455fae8cbf7f92838b15f59ae72 The scripting/workflow changes
2. 301a088057c92239aced7f92b60b6bb266275a17 The converted sources

PR #719, opened 2023-11-08
Version 2 import script/workflow

RickyDaMa:
Adds Harry's new Glyphs converting script, and rewrites the import workflow to make use of it

Aims to automate the process of #718


belluzj:
@MariannaPaszkowska this is ready to merge. It disables the overlap bit code in the `Makefile`, and the stat table in `config.yml`. We'll re-enable both of these later when we have the slanted sources.

We have checked that the name table matches, and to achieve that we needed a little bit of import script magic (renaming the default UFO to Regular). We also checked that the vertical metrics match, and they're all good, and also they will be checked by the Fontbakery QA that compares against Google Sans.

We've found other issues which can be fixed later by Octavio's team and have listed them in https://github.com/googlefonts/googlesans-flex/issues/727

In particular, the newly added brace layers will need to be reviewed to allow us to build the fonts. We've been able to make progress regardless by changing the brace layer locations ourselves to test the build + import but we did so without any consideration for Octavio's design intentions, so we would like for Octavio's team to do these fixes, so the layers can serve their intended purposes.

Issue #720, opened 2023-11-09
Update dependencies


Hoolean:
## Goals

- Update fontbakery, see:
  - #276
   - #716
- Use the latest toolchain for the v2 release

## Plan

1. Separate fontbakery from the rest of dependencies and always use the newest version:
    - Use f-actions/font-bakery for the CI (thanks @chrissimpkins 🤜🤛)
    - Use a separate venv in the local Makefile (or defer to the user-installed version?)
2. Update the build dependencies
3. Check failing tests
4. Regression test

RickyDaMa:
Resolved by #722 & #721

PR #722, opened 2023-11-10
Update build dependencies

Hoolean:
This PR updates build dependencies with the strategy in #720.

## TODOs

* [ ] Check for any broken workflows or differences in the built TTFs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,8 @@ venv: venv/touchfile
 
 build.stamp: venv sources/config.yaml $(SOURCES)
 	rm -rf fonts/
-	venv/bin/gftools builder sources/config.yaml
+	. venv/bin/activate \
+		&& gftools builder sources/config.yaml
 	# Font-v cannot deal with worktrees, which we use for imports. See
 	# https://github.com/source-foundry/font-v/issues/169. Just skip it.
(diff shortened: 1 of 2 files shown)


RickyDaMa:
Looking at the biggest VF (with slnt):
GDEF, GPOS, HVAR are very different
MVAR is almost entirely deleted with new deps
gvar has like 3 fewer entries in the new font
GSUB is semantically identical I think
name has only expected differences

Looking at the VFs and playing with sliders, I can't see any differences

Shall I run diffenator2 on the before/after big VFs?

Take a look at the TTFs and TTX tables yourself with the attached zip archives

[Google Sans Flex main.zip](https://github.com/googlefonts/googlesans-flex/files/13321421/Google.Sans.Flex.main.zip)

[Google Sans Flex new-deps.zip](https://github.com/googlefonts/googlesans-flex/files/13321427/Google.Sans.Flex.new-deps.zip)

RickyDaMa:
Ran diffenator2 (v0.3.2), reports can be downloaded [here](https://drive.google.com/file/d/1Yo-ULzrPKWUQj00yB_ljNQwP5dwbbqHA/view?usp=sharing) (too big for GitHub). I haven't looked at all of them, but the ones I did look at, I saw no differences between before & after

MariannaPaszkowska:
Thank you, all! 🚀

PR #723, opened 2023-11-15
Flatten UFO hierarchy and adopt new naming scheme before import

Hoolean:
This PR moves every UFO from their sub-directory into the route of the `sources/` folder, and renames them with the new 2-letter axes tags.

e.g.
* Old: `sources/italic/GoogleSansFlexItalic-opsz144-wdth100-wght1-ROND0.ufo`
* New: `sources/GoogleSansFlex-wg1-wd100-oz144-GD0-RD0-sl-10.ufo`

This is applied to the V1 sources, in order to make the diff from moving to V2 as manageable as possible.

## QA

A ttx-diff of the build from these sources against those on `main` showed no semantic differences (only checksum/modified time/commit version tag).

## Other changes

This PR also makes two other changes, neither of which affect the final built TTFs:

* Combines the regular and italic features into a single file, as at present they have no semantic difference
* Removes UFOs that are not referenced by any designspace, e.g. those in the `instances/` sub-directory or with `copy` in their name


belluzj:
@MariannaPaszkowska this is ready to merge as well

PR #724, opened 2023-11-16
Rename sources in v2 glyphspackage

belluzj:
@Hoolean @RickyDaMa Please test that importing from this branch does indeed emit UFOs with the new naming scheme. If validated, then please check with @octaviopardo when is a good for his team to merge the changes (or for us to re-run the rename script on top of new changes).


belluzj:
@MariannaPaszkowska @octaviopardo this is ready to merge; please note that the target branch is Octavio's branch. We're changing two things:
* rename masters to the new naming scheme
* add a Variable Font Origin to make the default axis positions match v1

The other changes that should happen on Octavio's side are the brace layer coordinates, which need some review to avoid duplicate brace layers for the same glyph. We believe it's better if the design team do these changes, as they would know what they're doing from a design point of view.

For reference this Glyphs.app macro will report duplicated brace layers:

```python
from collections import defaultdict

glyph_loc_to_layers = defaultdict(set)
	
for glyph in Font.glyphs:
	for layer in glyph.layers:
		coordinates = layer.attributes.get('coordinates')
		if coordinates is None:
			continue
		glyph_loc_to_layers[glyph.name, coordinates].add(layer)

for (name, coordinates), layers in glyph_loc_to_layers.items():
	if len(layers) <= 1:
		continue
	print("Name", name)
	print("Coordinates", coordinates)
	print("Layers", layers)
	print("Masters", [Font.masters[layer.associatedMasterId] for layer in layers])
	print()
```

PR #729, opened 2023-11-17
Temporarily strip brace layers on import

Hoolean:
This is necessary until we align brace layers at the same position to the same master in the upstream Glyphs sources, or we make glyphsLib more flexible (see googlefonts/glyphsLib#956).


Hoolean:
@MariannaPaszkowska Tested locally and brace layers are no longer present in the designspace, and so this is ready for review and merge; it will progress towards importing Octavio's latest sources on the CI 🚀

Issue #726, opened 2023-11-16
Preferred axis order in fvar table


MariannaPaszkowska:
What would be an ideal user-facing axis order in fvar table?

Hoolean:
Precedent:

* **Google Sans v11.001**
  * opsz, wght, GRAD, ital*
* **Google Sans Flex v1.002**
  * wght, wdth, opsz, ROND, slnt

*only present in STAT, as there is no combined upright+italic TTF

---

@MariannaPaszkowska We will be swapping the order of opsz and wght for #731; shall we adopt:

> opsz, wdth, wght, GRAD, ROND, slnt

...for our pre-release V2 sources to see how different programs and deployment packaging behave? :)

MariannaPaszkowska:
@Hoolean yes, let's try this out, many thanks!

Hoolean:
@MariannaPaszkowska @chrissimpkins We have confirmed that the new order is in effect [in main](https://github.com/googlefonts/googlesans-flex/actions/runs/7196079980) for both `fvar` and `STAT`, and so this is ready for any application or deployment testing we may wish to do :)

> opsz, wdth, wght, GRAD, ROND, slnt

PR #731, opened 2023-11-20
Reinstate STAT for main TTF, and add Grade axis bounds and labelling

Hoolean:
This PR reinstates the STAT table for the main TTF, which was temporarily disabled for the V2 import, and adds bounds (min/max) and labelling for the Grade axis that will be reintroduced in V2.

## TODOs

* [x] Test import script and config changes with V2 sources
* [x] Confirm axis bounds and labelling


Hoolean:
When testing this PR, we observed that the Flex build configuration declares 'opsz' before 'wght' in the STAT table axis order, but that this is not observed in the final fonts:

- **Google Sans v11.001**
  - `fvar`: opsz before wght
  - `STAT`: opsz before wght
- **Google Sans Flex v1.002**
  - `fvar`: wght before opsz
  - `STAT`: wght before opsz
    - gftools [defines the opposite order](https://github.com/googlefonts/googlesans-flex/blob/d3cb8008/sources/config.yaml#L5-L10) but this is not used in the final fonts

Having discussed, for Flex V2 we want to test with 'opsz' before 'wght' in both `fvar` and `STAT`, to match the original Google Sans:
- [x] Move 'opsz' before 'wght' in `fvar`
- [x] Move 'opsz' before 'wght' in `STAT` (investigate gftools behaviour above)
- [ ] Test how this affects the built font in editors and deployment

Hoolean:
We have now confirmed in testing that this PR reinstates the STAT table from V1 with the V2 sources, with the addition of the Grade axis, and following the new fvar order we are testing for #726.

Keeping #726 open to track further fvar and STAT testing, this PR is ready for review and merge immediately after we merge the first V2 import into main 🚀

MariannaPaszkowska:
Thank you, great job!

PR #742, opened 2023-12-05
Fix Makefile to fix Import action

Hoolean:
This PR fixes the Makefile issue that caused [this Import action run](https://github.com/googlefonts/googlesans-flex/actions/runs/7104280354/job/19338879912#step:10:3353) to fail 🔧


Hoolean:
@MariannaPaszkowska Ready for review and merge :)

Issue #735, opened 2023-12-01
Instance labelling difference between 'fvar' and 'STAT'


Hoolean:
For V1 and V2 (as of #731), there is a labelling difference between:

* Applications that label fvar instances by their fvar label; and
* Applications that label fvar instances by their positions in STAT.

Is there anything we want to change here? We may also have control over ordering, although this is more difficult to enforce consistently across applications.

## Example

| ![fvar instances](https://github.com/googlefonts/googlesans-flex/assets/2150762/bfd43f87-c2f8-4922-a350-eef949f6bd28) | ![STAT values](https://github.com/googlefonts/googlesans-flex/assets/2150762/fe834e3b-49dd-4572-9a68-ba1e7f31682f) |
|-|-|
| **fvar instance labels** | **STAT-derived labels** |

Hoolean:
TODOs from discussion earlier:

* [ ] Rename "Slanted" at `STAT` table axis extreme to "Italic", to match the `fvar` instance names
* [x] Keep "Slant" for name of the axis

PR #748, opened 2023-12-08
Sub VFs approach 1: slice from the largest VF

Hoolean:
_(1/2) For testing #747_

This PR builds upright-only and italic-only VFs by slicing from the largest VF post-compilation.

## Pros

* Avoids redundancy in sources
* Avoids duplicated work in compilation

## Cons

* Less source-level control of sub-VFs, may require post-build TTF adjustments

## TODOs

* [ ] Diff key tables with corresponding V1 VFs
* [ ] Review: this PR [changes the STAT label](https://github.com/googlefonts/googlesans-flex/commit/5263a77f) for `slnt=-10` from "Slanted" to "Italic" to encourage the subsetter to use italic-appropriate bits and naming
    * Is this change desirable? (see #735 for other considerations)


anthrotype:
> Less source-level control of sub-VFs, may require post-build TTF adjustments

the new public.fontInfo may help there (if we implement in ufo2ft etc.)

Hoolean:
Tidying old branches and so closing this PR: we proceeded with a single VF, and there are no workarounds we need to preserve from here.

PR #749, opened 2023-12-08
Sub VFs approach 2: build from separate designspaces

Hoolean:
_(2/2) For testing #747_

This PR builds upright-only and italic-only VFs by generating and building from three separate but overlapping designspaces.

## Pros

* Full source-level control of sub-VFs
* Potential to transition to single DS5 designspace later

## Cons

* Duplicated work in compilation
* Currently requires three redundant designspaces until compilation pipeline has DS5 support
* More complicated import process

## TODOs

* [ ] Diff key tables with corresponding V1 VFs
* [ ] If chosen: integrate into Glyphs 3 import process


belluzj:
Closed in favour of https://github.com/googlefonts/googlesans-flex/pull/748

PR #754, opened 2023-12-13
Change label of slnt extreme to Italic, for consistency with fvar

Hoolean:
See #735

### DIFF
--- a/sources/config.yaml
+++ b/sources/config.yaml
@@ -161,7 +161,7 @@ stat:
          value: 0
          linkedValue: -10
        - flags: 0
-         name: Slanted
+         name: Italic
          value: -10
 # TODO: Recreate based on new STAT values above.
 #  "GoogleSansFlex[ROND,opsz,wdth,wght].ttf":


Hoolean:
## New STAT

* The label of `slnt=-10` is now "Italic" instead of "Slanted"
* The name of the axis is still "Slant"

## Diff

|  | **fvar instance labels** | **STAT-derived labels** |
|-|-|-|
| Before | Instance dropdown, labelled by fvar | Instance dropdown, labelled by STAT |
| After | Instance dropdown, labelled by fvar | Instance dropdown, labelled by STAT |

Hoolean:
@MariannaPaszkowska @chrissimpkins Ready for review; does the new behaviour look good, and are there any other changes we would like to make to the STAT labelling before we merge the PR?

Hoolean:
For caveats, this new STAT mixes 'Slant' and 'Italic' in the same context:

- 'Slant' as the name of an axis; and
- 'Italic' as a position on that axis.

We would expect them to be surfaced together in a GUI that includes e.g. an axis slider with labelled ticks, but we are not immediately aware of any software that does that; this may just be a semantic compromise necessary to get the axis and instance labelling we want elsewhere.

chrissimpkins:
I don't think that our internal version of fonts.google.com where we host the brand families uses STAT for any UI.  Most of our designers use Figma and I believe that is based solely on fvar.   Probably safe but will let you know if we hear about any issues as we roll it out.  Ty!

MariannaPaszkowska:
@Hoolean All looks good, many thanks!

Hoolean:

Hello both! It looks like this introduced [a FontBakery FAIL](https://github.com/googlefonts/googlesans-flex/actions/runs/7247658807/job/19742035705#step:8:1272), as there is a check that asserts that 'Italic' should only appear in an `ital` axis.

There is discussion upstream on whether `slnt` is a valid exception (fonttools/fontbakery#4199), which relates the labelling [to Google Fonts deployment](https://googlefonts.github.io/gf-guide/variable.html#ital--slnt) (and may also have relevance for when we decide whether to build upright-only and italic-only VFs, see #749). Is this relevant to Flex's deployment too? :)

PR #761, opened 2023-12-18
Remove older ROND reflow check

Hoolean:
The upstream FontBakery GRAD reflow check has been adapted to check for ROND reflow too, and so this project-specific check is now redundant.

* Upstream issue tracking feature:
  * fonttools/fontbakery#4200
* Related change in GS:
  * googlefonts/googlesans#584


Hoolean:
The old and new tests are based on [the same original code](https://github.com/fonttools/fontbakery/blob/2e94360f/Lib/fontbakery/profiles/googlefonts.py#L6062-L6152), and [the output of the two tests in the Beta release](https://github.com/googlefonts/googlesans-flex/actions/runs/7251058830/job/19752600067#step:8:1829) is identical up until it is truncated¹, and so this PR is ready for review and merge 🚀 


---

¹
image

PR #763, opened 2023-12-18
Small tidy: remove deleted designspaces from Makefile

Hoolean:
Spotted by Ricky when simplifying the workflows: this was silently ignored but better not to keep around.

cc @RickyDaMa


Hoolean:
Ready for review and merge; when testing locally, FontBakery shows the same number of tests of each status before and after this change 🎉

RickyDaMa (approved):

PR #765, opened 2023-12-20
Fix FontBakery false-positives and ERRORs

Hoolean:
The V2 addition of the GRAD axis (with accompanying file name changes) and temporary removal of features led to some false-positives and ERRORs from our checks.

This PR updates check expectations and implementations to replace each with their correct status.

## New expectations

The `GRAD` axis is now formally expected by the tests to have the name `Grade` and to have a default value of `0`.

## Additional changes

This PR also promotes unexpected or misnamed axes from WARN to FAIL, to ensure we do not ship regressions in this area.


Hoolean:
(the false-positives and ERRORs affect engineering checks and not design checks, and so there is no urgency to merge before the end of the holidays ⛄)

RickyDaMa (approved):

Issue #854, opened 2024-02-13
Import workflow: FB issues


MariannaPaszkowska:
Hello team, something is wrong with our FB test suite. The latest import took 1h 15m and FB didn't finish running: 

https://github.com/googlefonts/googlesans-flex/actions/runs/7877303401/job/21493644913 

Could you please look into that?

We get an error in  
```
 >> Check:   com.google.fonts/check/glyphsets/shape_languages
    Desc:    Shapes languages in all GF glyphsets.
    Files:   GoogleSansFlex[GRAD,ROND,opsz,slnt,wdth,wght].ttf

    ERROR Failed with ImportError: cannot import name                           
          'unicodes_per_glyphset' from 'glyphsets.definitions'                  
          (/home/runner/work/googlesans-flex/googlesans-flex/venv_bakery/lib/pyt
          hon3.11/site-packages/glyphsets/definitions/init.py)                  
          ↳ File                                                                
          "/home/runner/work/googlesans-flex/googlesans-flex/venv_bakery/lib/pyt
          hon3.11/site-packages/fontbakery/checkrunner.py", line 183, in        
          _exec_check                                                           
          for sub_result in result:  # Might raise.                             
          ↳ File                                                                
          "/home/runner/work/googlesans-flex/googlesans-flex/venv_bakery/lib/pyt
          hon3.11/site-packages/fontbakery/profiles/googlefonts.py", line 3494, 
          in com_google_fonts_check_glyphsets_shape_languages                   
          glyphsets_fulfilled = get_glyphsets_fulfilled(ttFont)                 
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                       
          ↳ File                                                                
          "/home/runner/work/googlesans-flex/googlesans-flex/venv_bakery/lib/pyt
          hon3.11/site-packages/fontbakery/profiles/googlefonts_conditions.py", 
          line 739, in get_glyphsets_fulfilled                                  
          from glyphsets.definitions import unicodes_per_glyphset,              
          glyphset_definitions                                                  
    Result: ERROR
```
and the last fail seems to be 
```
 >> Check:   com.google.fonts/check/varfont/duplexed_axis_reflow
    Desc:    Ensure VFs with duplexed axes do not vary horizontal advance.
    Files:   GoogleSansFlex[GRAD,ROND,opsz,slnt,wdth,wght].ttf

     FAIL The following glyphs have variation in horizontal                     
[...]

PR #856, opened 2024-02-13
Fix Fontbakery hang & reduce max job durations

RickyDaMa:
Related to #854 

Follow-up with better fix to come, issue is likely caused by pinning only Fontbakery and not its dependencies


MariannaPaszkowska:
Many thanks!

PR #893, opened 2024-02-29
Reduce log spam from `com.google.fonts/check/googlesansflex/sources/suspicious_kerning_values`

RickyDaMa:
Use Fontbakery's built-in list utility so that it limits the number of items printed by default. This should massively reduce the ~12k log lines we've been seeing of late


RickyDaMa:
Down from ~12k log lines to ~4k!

PR #919, opened 2024-04-03
Only build PRs that target main

Hoolean:
Update(Ricky): changed approach to check if a PR is targetting `main`

---

This PR stops our automatic build workflow from running for PRs that only change the sources within `sources/design-source/`.

Changes in this folder do not affect the UFO sources or pipeline until manually imported, and so have no effect on the final TTFs.

## Workflows

This would avoid booting the CI for these recent builds (highlighted):

List of recent build workflows, with those that would no longer be issued highlighted

## Future

If we find that CI feedback for Glyphs PRs would be helpful, we could reinstate automated builds but with the "Build from Glyphs package" workflow instead, depending on CI minutes available 🚀

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,14 +8,11 @@ on:
         required: true
   push:
     branches:
-      - main
+    - main
   pull_request:
     types: [opened, synchronize, reopened]
-    # Do not build on PRs that only change the design sources; these do not
-    # affect the UFO sources or pipeline until manually imported, and so should
-    # have no effect on the final TTFs.
-    paths-ignore:
-      - 'sources/design-source/**'
+    branches:
+    - main
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


Hoolean:
After this is merged we can make a test PR changing only the Glyphs sources to confirm that a build job is no longer started; if it sneaks through anyway, we can cancel it on the spot and try a different approach 💥

Hoolean:
Extra note: if this commit works, we might need to cherry pick it into the branches where design work is happening :)

.github/workflows/build.yaml
```diff
@@ -11,6 +11,11 @@ on:
       - main
   pull_request:
     types: [opened, synchronize, reopened]
+    # Do not build on PRs that only change the design sources; these do not
+    # affect the UFO sources or pipeline until manually imported, and so should
+    # have no effect on the final TTFs.
+    paths-ignore:
+      - 'sources/design-source/**'
```
RickyDaMa: From my experience, this won't work due to how many files there are in the Flex repo :(

Might need to go for a naming convention approach for branches we want to exclude

RickyDaMa (changes requested):

RickyDaMa (approved):

PR #928, opened 2024-04-11
Rewrite diffenator2 workflow

RickyDaMa:
Based on GS Mono: https://github.com/googlefonts/googlesans-mono/blob/5cf114eb8a5a8314de39f64d40638dc7c4dab986/.github/workflows/diffenator2.yml

Let's you compare two arbitrary points in Git history without requiring special branch names

Due to https://github.com/googlefonts/diffenator2/issues/134, the current workflow doesn't work any more now that our releases have statics

### DIFF
--- a/.github/workflows/diffenator.yaml
+++ b/.github/workflows/diffenator.yaml
@@ -1,83 +1,98 @@
-name: diffenator2 tests
+name: Diffenate builds
 
 on:
-  push:
-    branches:
-      - "diff-*"
-
+  workflow_dispatch:
+    inputs:
+      before:
+        description: The 'before' git reference (tag, branch, or commit)
+        required: true
+      after:
+        description: The 'after' git reference (tag, branch, or commit)
+        required: true
 
 jobs:
-  build:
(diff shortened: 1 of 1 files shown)


m4rc1e (approved):
LGTM

I've never used a workflow dispatch! This looks great.

I haven't checked for typos etc but I'm sure these can be fixed once merged. Personally, I've never found a good way to test github actions. I usually just test run them on a fork until it works, then merge all the testing commits together.

RickyDaMa:
> Personally, I've never found a good way to test github actions.

Yep, same. We (Dalton Maag) have private forks of the big Google repos; I've tested this workflow here and it was fine in a basic smoke check ✅

m4rc1e:
I'm on holiday next week but when I come back, I'll alter the diffenator2 github action so you can specify which archive to use in a release.

RickyDaMa:
Enjoy your break!

No rush on that feature - I think we should stick with the approach in this PR in any case because then we're not limited to only testing against the previous release. Ultimately though, that's for Marianna to decide on

Issue #936, opened 2024-04-29
dollar and cent: small design issues


belluzj:
/dollar: the gap between the vertical bars shows up at location `'opsz' 6, 'wdth' 151, 'wght' 1, 'GRAD' 100, 'ROND' 100, 'slnt' -10` along the weight axis

![gsflex_dollar_gap_bug](https://github.com/googlefonts/googlesans-flex/assets/3616772/be29a78a-4c11-4e75-81c9-62df94ac8115)

---

/cent: the spacing is strange at location  `'opsz' 144, 'wdth' 25, 'wght' 400, 'GRAD' 0, 'ROND' 0, 'slnt' -10` along the weight axis

![gsflex_cent_width_bug](https://github.com/googlefonts/googlesans-flex/assets/3616772/2f65b14f-a5ae-4514-9002-8c7d929c20ad)

octaviopardo:
@MariannaPaszkowska  
Fixed in branch "main-+-dollar-&-cent-fix"
Images attached (top new build, bottom v2.1)
@belluzj 1 beer for you for the catch

![Captura de pantalla 2024-05-02 a las 13 04 44](https://github.com/googlefonts/googlesans-flex/assets/2181684/01874bcf-d657-41a6-95c8-4bd6f601143c)
![Captura de pantalla 2024-05-02 a las 13 04 00](https://github.com/googlefonts/googlesans-flex/assets/2181684/fd09fc7e-b3f8-4741-b4f5-5a0960794e1f)

PR #943, opened 2024-05-13
fixes issue in dollar and cent (Issue #936)

octaviopardo:


MariannaPaszkowska:
Hi @octaviopardo I'm reviewing this PR and I see two new encoded glyphs
̵
̷

are those intentional or is this caused by that old issue where non-exporting glyphs sneak in? 
cc @belluzj @RickyDaMa

octaviopardo:
Both glyphs are set to non-export

Captura de pantalla 2024-05-14 a las 18 11 08

MariannaPaszkowska:
@RickyDaMa @belluzj Did we unintentionally change something in the latest update? Non-exporting glyphs are in the build fonts now.

RickyDaMa:
Last I recall, we haven't done an import since fixing the issue, so the UFO sources will still be built with all glyphs, not respecting those marked as excluded in the design/Glyphs sources

In any case, we'll need to run the import workflow on the changes in this PR, so then everything should come out in the wash

MariannaPaszkowska:
Of course! I forgot to import. That was silly of me! Thanks a lot @RickyDaMa !

MariannaPaszkowska:
fixed in import #947

PR #935, opened 2024-04-19
Workspace VFs refactor

RickyDaMa:
Will close #932 

Changes workspace fonts from being statics to being VFs on the weight axis, with a reduced range of 100..=900, instead of 1..=1000

⚠ Note: this branch is based on top of the dependency updates, as with current dependencies the font will not visually change as the weight value changes. ⚠ (Specifically, the latest version of `fonttools` fixes this issue)

- [x] Update cut_instances.py
- [x] Check updated fonts pass Fontbakery
- [x] Add README for workspace archive
- [x] Update & test GitHub workflow for release

PR #891, opened 2024-02-26
Unpin Fontbakery & update profiles for 0.12

RickyDaMa:
So long as the pipeline passes for this PR we're safe to be back on the latest & greatest Fontbakery again 😄

PR #952, opened 2024-05-22
Fix typo in Fontbakery check code + expand Workspace fonts README

belluzj:
There was a typo in the FontBakery check code, resulting in the following failure:

image


RickyDaMa (approved):
Shoot sorry, must have messed that up when I was rebasing #891 onto #935! Good find

PR #960, opened 2024-07-09
Switch to free runners and increase timeouts

RickyDaMa:
For more information see #961 

Faster runners are currently unavailable

This configuration is known to work fine on the testing repo, so should work for us for now


RickyDaMa:
Superceded by #962

PR #980, opened 2024-07-18
Include `com.google.fonts/check/gsub/smallcaps_before_ligatures`

### LINKED ISSUE #976: Re-include `com.google.fonts/check/gsub/smallcaps_before_ligatures`
Context: #972

Summary: check isn't in a great state and we're about to release, so disable it for now and the next Fontbakery release after v0.12.8 will be fine as the improvement is already merged into `main`

To-do: revert #977

RickyDaMa:
Closes #976 - a new Fontbakery release has been cut, addressing the original concerns that resulting in this check being excluded

### DIFF
--- a/qa/check-sources.py
+++ b/qa/check-sources.py
@@ -26,8 +26,4 @@ PROFILE = {
             "com.google.fonts/check/googlesansflex/sources/no_open_corners",
         ]
     },
-    "exclude_checks": [
-        # https://github.com/googlefonts/googlesans-flex/issues/972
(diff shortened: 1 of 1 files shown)


MariannaPaszkowska:
@RickyDaMa Is there any work left here?

RickyDaMa:
Yes there is, please merge this when you get a minute! This PR re-includes a check we excluded at one point due to a bug (which has now been fixed, hence re-including it)

PR #989, opened 2024-07-20
Remove unaccessible kerning

schriftgestalt:
This kerning was on masters that have a Link Metrics parameter but still have own kerning defined (that is never visible in the UI). This is probably some leftover from previous operations.


Hoolean:
The latest commit should fix the CI job hanging, although it also brings in an earlier change we made to only run the UFO build job on PRs that target `main` (#919), and so it is expected that no job is run at all now.

Issue #981, opened 2024-07-18
v2.002 vs upcoming v2.003 diffenator2 diffs


m4rc1e:
I've just compared the main branch against the old v2.002 release. Are we expecting such significant spacing changes for such a minor release?

![d](https://github.com/user-attachments/assets/fcf4e72f-bc8d-41e1-9ec3-ea7cc6f16ec4)

Diffenator2 command used is 

```BASH
diffenator2 diff -fb /Users/marcfoley/Downloads/GoogleSansFlex\[GRAD\,ROND\,opsz\,slnt\,wdth\,wght\].ttf -fa fonts/variable/GoogleSansFlex\[GRAD\,ROND\,opsz\,slnt\,wdth\,wght\].ttf -o ~/Desktop/gs-flex2 --styles cross_product
```

m4rc1e:
I'm not just checking the masters and there seems to be a lot of differences

![dd](https://github.com/user-attachments/assets/36349a10-0382-46d7-a079-b26f3ccdc66e)

The ligs in Thin weights seem more compact.

m4rc1e:
![ddd](https://github.com/user-attachments/assets/1da48b88-658e-4831-9a4b-314ffd6fd465)

m4rc1e:
Just some basic stats comparing v2.002 against current main:

Kerning has changed for 215 masters.
Over 58,000 ufo glyphs have been modified (mainly changing point type etc)

MariannaPaszkowska:
thanks @belluzj

chrissimpkins:
Are we sure that the regression is v2.002 to v2.003 and that we didn't have spacing regressions at some stage between our latest release v2.000 and v2.002?  These spacing changes are all regressions vs. our v2.000 release?

MariannaPaszkowska:
@chrissimpkins I can confirm that those changes were absent in v2.001 and 2.002.

MariannaPaszkowska:
#996 adds context to this.

PR #990, opened 2024-07-22
Import from GSF-Ashler-v2.2-OctDev

MariannaPaszkowska:
GSF-Ashler-v2.2-OctDev at 1308f07d17


MariannaPaszkowska:
will close #981

PR #1010, opened 2024-07-31
Check vertical extremes

Hoolean:
This branch contains:

* The script used to generate [our report on the y-axis extremes in each source](https://docs.google.com/spreadsheets/d/1ISNGNcIr-gFNhjCZUb2AmZ9N6aRNe2LQYQofAshdogw/edit?usp=sharing)
* A FontBakery check for this project's OS/2.WinAscent and OS/2.WinDescent semantics

The latter is WIP, and in general this branch is non-urgent and does not need merging until after the v2.003 release :)

### DIFF
--- a/qa/checks/googlesans.py
+++ b/qa/checks/googlesans.py
@@ -25,6 +25,8 @@ from fontbakery.utils import (
     compute_unicoderange_bits,
     unicoderange_bit_name,
 )
+from fontTools.pens.boundsPen import BoundsPen
+from fontTools.ttLib import TTFont
 
 # Each VF we build will have one of these suffixes depending on whether it
 # includes the full designspace or is restricted to upright or italic only.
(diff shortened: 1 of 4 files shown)


chrissimpkins:
Does this implementation cover *shaped* y-min/max extremes?

chrissimpkins:
Seems ok to merge?  Is there anything outstanding in this PR at this stage?

qa/checks/googlesans.py
```diff
@@ -298,3 +300,86 @@ def com_google_fonts_check_googlesansflex_variable_fvar_default(font: Font, ttFo
...
+        variated_tags = {
+            tag for record in mvar.table.ValueRecord for tag in record.ValueTag
+        }
+
+        if "hcla" in variated_tags:
+            yield FAIL, "TTF varies OS/2.usWinAscent across its designspace"
+        if "hcld" in variated_tags:
+            yield FAIL, "TTF varies OS/2.usWinDescent across its designspace"
+
+    # TODO: This is only named instances; we should include everywhere we have
```
RickyDaMa, line 347: Still outstanding

qa/checks/googlesans.py
```diff
@@ -298,3 +300,86 @@ def com_google_fonts_check_googlesansflex_variable_fvar_default(font: Font, ttFo
...
+            (_, y_min, _, y_max) = bounds_pen.bounds
+
+            smallest_y_min = (
+                y_min if smallest_y_min is None else min(y_min, smallest_y_min)
+            )
+            largest_y_max = (
+                y_max if largest_y_max is None else max(y_max, largest_y_max)
+            )
+
+    # TODO: Round floats in some direction (but consider how this will interact with the +1).
```
RickyDaMa, line 380: Still outstanding

RickyDaMa (commented):
These should probably be addressed before merging

Also, we may want to revise the approach (e.g. looking at shaped strings) based on the discoveries of the Pixel fallback stack saga

MariannaPaszkowska:
@Hoolean "It's a good idea to wait till after the strategy for the Android subsets is defined."

belluzj:
@chrissimpkins should we close this PR now that we've decided not to subset GS Flex? Thanks in advance

belluzj:
Chris: not related to GSF subsetting, so we should keep.
Chris: intent for Win metrics: every time we add glyphs we should adjust the win metrics (the values are entered by hand). It's not a novel approach, we've been doing that for Google Sans and we should keep doing that for all the brand families

RickyDaMa:
> Does this implementation cover _shaped_ y-min/max extremes?

No, it just checks glyphs individually

RickyDaMa:
Approach taken in this PR to establish extremes is not the approach we want based on https://github.com/googlefonts/googlesans-code/issues/56#issuecomment-2531718958

Closing but not deleting branch as there may be some useful code here for future developments

PR #1013, opened 2024-08-05
Update build dependencies

RickyDaMa:
Updates build dependencies to the latest at the time of commit. I was going to update Python to 3.12, but it seems to significantly increase built & test times

Most recent diffenator run: [here](https://github.com/googlefonts/googlesans-flex/actions/runs/10268362064) - no changes!


RickyDaMa:
Both builds and tests now seem to be taking longer 🤔 

I've increased the timeout for tests, but this'll only kick in after the PR is merged due to how the test workflow works

RickyDaMa:
Tried to reproduce the performance difference locally, failed:

image

<details>
<summary>hyperfine invocation used</summary>

```shell
hyperfine --warmup 1 --runs 4 \
  -n "Python 3.11 old deps" "venv/bin/gftools builder sources/config-old.yaml" \
  -n "Python 3.11 new deps" "venv311-new-deps/bin/gftools builder sources/config.yaml" \
  -n "Python 3.12 new deps" "venv312/bin/gftools builder sources/config.yaml"
```

</details>

MariannaPaszkowska:
let's stick to python 3.11 and let's review if the PR reflects this. cc @RickyDaMa

RickyDaMa:
PR is all good and updated, only failing the pipeline because of #1016

MariannaPaszkowska:
I confirm no differences. Many thanks!

Issue #1016, opened 2024-08-19
Language coverage: new errors about Igbo & Yoruba


RickyDaMa:
From [here](https://github.com/googlefonts/googlesans-flex/actions/runs/10183450341/job/28815724080#step:10:1118):

```
Igbo:
    PASS: All base glyphs were present in the font
    PASS: All mark glyphs were present in the font
    PASS: All auxiliary glyphs were present in the font
    PASS: No unattached mark glyphs were produced shaping the text 'A Ụ T SH ṅ N NW L m ẹ ch sh nw Ị GH i H M z ụ gb d KW f ny g w o E P B I kw u r K p gh gw kp GB v U ọ j k W l t Y J KP a n R S e G h D b y ị s V Ṅ Ẹ F NY CH Ọ GW O Z'
    PASS: No unattached mark glyphs were produced shaping the text 'ị́Ị́ị̀Ị̀m̀M̀ọ́Ọ́ọ̀Ọ̀ụ́Ụ́ụ̀Ụ̀'
    WARN: No variant glyphs were found for .notdef
  PASS Hausa
  PASS Vietnamese
  Yoruba:
    FAIL: Some base glyphs were missing: Ɔ̀, Ɔ́, Ɛ, Ɛ̀, Ɛ́, ɔ̀, ɔ́, ɛ, ɛ̀, ɛ́
    PASS: All mark glyphs were present in the font
    WARN: Some auxiliary glyphs were missing: Ɔ̀, Ɔ́, Ɛ, Ɛ̀, Ɛ́, ɔ̀, ɔ́, ɛ, ɛ̀, ɛ́
    FAIL: Shaper produced a .notdef
    FAIL: Shaper produced a .notdef
```

Currently unsure what's changed to cause the regression here (no source changes; no dependency changes that should matter), but Octavio's confirmed that the flagged problems are present in the sources, so we should probably work on fixing them.

RickyDaMa:
From Octavio:

> the problem is that those glyphs were not requested for v2.3, and I'm not 100% sure if they will be requested for 3.0 so there is no way for me to fix that. Are you sure that in the previous round shaperglot was checking them?? Maybe it was just checking a partial set of Yoruba?

chrissimpkins:
**Igbo**: _WARN no variant glyphs were found for .notdef_

what does this mean?

chrissimpkins:
[...]

PR #1017, opened 2024-08-26
Add script that reports each glyph's size contribution to 'gvar', to find optimisations

Hoolean:
This draft PR contains the raw script used to produce the raw data of the report breaking down [`gvar` size by glyph](https://docs.google.com/spreadsheets/d/1hn46LgquIAnpYQ8s-PacGnAFv0ikveBwKqAPPP0VcSQ/edit?usp=sharing).

Before merging, we should adapt this PR to run the script on our build workflows - it only takes a second or so - and output something closer to the full report above as an uploaded CSV and/or table in the build log.

---

We can also think about where it could have a nice OSS home to make it available to hand for other projects :)

e.g. fonttools, [font-size](https://github.com/source-foundry/font-size), Font Bakery?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,13 @@ font-size: venv build
 	-@[ -n "${GITHUB_RUN_ID}" ] && echo "::group::Install font-size"
 	venv/bin/pip install -U font-size
 	-@[ -n "${GITHUB_RUN_ID}" ] && echo "::endgroup::"
+	-@[ -n "${GITHUB_RUN_ID}" ] && echo "::group::TTF size report, by table"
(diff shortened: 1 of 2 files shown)


RickyDaMa (approved):

Hoolean:
@MariannaPaszkowska Hi Marianna! :)

This branch is ready to merge; when it is on `main`, every test workflow will include a report showing how much each glyph contributes to the `gvar` table, similarly to what the spreadsheet shows now.

(the CI fail on this branch is because of the shaperglot issued tracked at #1016 and is unrelated)

Issue #1021, opened 2024-08-28
Update the sources


MariannaPaszkowska:
It seems that the sources in the main are not v 2.003. Let's update that and see if we can automate that process.

RickyDaMa:
Would you have it be part of the import workflow? So that the design sources that are being imported from are updated as part of the PR

I've been wanting this for a while and it shouldn't be that bad to add

MariannaPaszkowska:
@RickyDaMa yes, that would be great!

RickyDaMa:
See #1023 :)

PR #1022, opened 2024-08-29
update design source to 2.003

MariannaPaszkowska:


MariannaPaszkowska:
partly resolves #1021

PR #1025, opened 2024-09-02
Add source checks for common automatic decomposition causes

### LINKED ISSUE #1019: Add automatic checks for automatic decomposition
This is a production/QA companion issue for #1018 to brainstorm how we can use Font Bakery to catch any automatic decomposition that may sneak into the project.

## Ideas

Source checks:

 * Check for any glyphs that are on the 'skip' list but which are referenced by unskipped glyphs as a component more than once
   * Would catch: /f_f_j, /f_f_b, etc
 * Check for any glyphs that apply near-identity transforms to their components
   * Would catch: /sterling.tf, etc

TTF checks:

 * Check for any glyph that has been automatically decomposed and is not in a configurable list of 'allowed' glyphs
   * Would catch: anything else, but not /questiondown and other necessary decompositions

Hoolean:
Checks whether sources:

- Use 'skipped' glyphs as components more than once
- Mix contours and components in a single glyph
- Apply a varying transform to a component across a designspace

Partial implementation of #1019 (TTF checks still needed).

### DIFF
--- a/qa/checks/sources.py
+++ b/qa/checks/sources.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+from collections import Counter
 from typing import Iterable, Tuple, Union
 
 from fontbakery import utils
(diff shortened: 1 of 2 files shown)


Hoolean:
Preview (partial lists):

Font Bakery FAIL showing list of glyphs using skipped components

---

Font Bakery FAIL showing list of glyphs using varying transforms

RickyDaMa (approved):
Looks good, but any reason why this couldn't be contributed to upstream Fontbakery? Only thing you'd need to change is to yield full `Message`s

qa/checks/sources.py
```diff
@@ -267,3 +268,141 @@ def check_no_open_corners(config, ufo: Ufo) -> CheckStatus:
...
+                uses_contours.add(glyph.name)
+            if glyph.components:
+                uses_components.add(glyph.name)
+
+    # Flag any glyphs that use both.
+    uses_mix = uses_contours & uses_components
+
+    if uses_mix:
+        yield (
+            FAIL,
```
RickyDaMa, line 354: Maybe this should just be a `WARN`?

qa/checks/sources.py
```diff
@@ -267,3 +268,141 @@ def check_no_open_corners(config, ufo: Ufo) -> CheckStatus:
...
+                assert isinstance(source.name, str)
+                seen_in.add(source.name)
+
+    # Fail for any component that has different transforms in different areas of
+    # the designspace.
+    for glyph, glyph_transforms in transforms.items():
+        for (index, component), comp_transforms in glyph_transforms.items():
+            if len(comp_transforms) > 1:
+                yield (
+                    FAIL,
```
RickyDaMa: This definitely should only be a `WARN`, maybe even `INFO` imo (though we filter out INFOs in our Fontbakery config, but if going upstream I'd say `INFO`)

qa/checks/sources.py
```diff
@@ -267,3 +268,141 @@ def check_no_open_corners(config, ufo: Ufo) -> CheckStatus:
...
+
+    # Flag any component that is skipped and would be decomposed into its parent
+    # glyphs more than once.
+    for component, counts in sorted(used_in.items()):
+        total = sum(counts.values())
+
+        if component in skipped and total > 1:
+            yield (
+                FAIL,
+                f"Component `{component}` is referenced at least {total} times, but is on the skip list, so will decompose :\n\n"
```
RickyDaMa: Nit: space before colon ;)

RickyDaMa (commented):
A couple extra royalty-free thoughts

qa/checks/sources.py
Hoolean: 🤮

Hoolean: Nice spot, thank you

qa/checks/sources.py
Hoolean: For this and the other note on severity, I think using `FAIL` is right for Flex as we ideally want to hit zero unnecessary automatic decompositions

Hoolean: If we find any that are indeed necessary then we can add the logic for those edge cases to the test, although right now I think it's okay; maybe before we merge we can check if we aware of any deliberately varying transformations?

Hoolean: In full agreement that we should nerf the severity if we upstream though; in that scenario I think it would make most sense to make these `WARN` and have a Flex-specific override to convert them into `FAIL`s

qa/checks/sources.py
Hoolean: (answered below 🤜🤛)

qa/checks/sources.py
MariannaPaszkowska: Yes, let's do FAIL for Flex and if upstream then probably it should be WARN

Hoolean: (🔀 after discussing with Octavio I have made this one a `WARN`, as it is a necessary design decision in some places - what we discussed still stands for the other checks though, and so I have kept those as `FAIL`)

RickyDaMa: Could be exclude some glyphs from the check somehow, save having noise in our Fontbakery reports?

Hoolean:
@MariannaPaszkowska This is now ready to merge :)

(the CI is failing because of real issues that the new file size checks are identifying, so no blocker there 🎉)

Hoolean:
- [ ] Check if this runs on Octavio's branch automatically after merging, or if we need to rebase

Hoolean:
Latest report: [fontbakery-sources-report.zip](https://github.com/user-attachments/files/16872749/fontbakery-sources-report.zip)

qa/checks/sources.py
```diff
@@ -267,3 +268,141 @@ def check_no_open_corners(config, ufo: Ufo) -> CheckStatus:
...
+
+        # Get layer from source definition.
+        layer = (
+            source.font.layers.defaultLayer
+            if source.layerName is None
+            else source.font.layers[source.layerName]
+        )
+
+        # Count components in each glyph.
+        for glyph in layer:
```
Hoolean, line 296: Skip exported glyphs when making counts here

Hoolean:
Update on upstreaming: we will likely upstream this as it would be useful for the original Google Sans, but let's merge here now to make it available for Flex right away

PR #1026, opened 2024-09-04
Build from Glyphs package: make tests opt-in instead of always enabled

RickyDaMa:
Does what it says on the tin, won't show up until merged to main

Doesn't require any updates to Octavio's branches

image


MariannaPaszkowska:
Excellent work, many thanks!

PR #1032, opened 2024-09-10
Run Fontbakery before Shaperglot, and still run both if either failed

### LINKED ISSUE #1031: Run shaperglot after FB tests
I think it would be good to could run shaperglot tests after the FontBakery and not before as it is set up now. This is because IMO it's likely we will need to test the font with FB while the character set is still being finished.

Hoolean:
Closes #1031, and also runs Shaperglot even if Font Bakery failed, as we discussed that this would be useful while having minimal CI cost (12 seconds on `main`)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -63,11 +63,6 @@ jobs:
     - name: Analyze TTF file size
       shell: bash
       run: make font-size
-    - name: Run shaperglot
-      shell: bash
-      run: |
-        touch build.stamp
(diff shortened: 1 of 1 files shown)


Hoolean:
PR is a draft, pending testing

Hoolean:
Confirmed that:

- Fontbakery and Shaperglot both run in the new order
- Shaperglot runs even when Fontbakery fails
- The action still fails when either Shaperglot or Fontbakery fails

Ready for review :shipit:

PR #1043, opened 2024-09-17
Import, but with /dollar from v2.003

Hoolean:
This PR contains the changes from the last test import (`import-GSF-Ashler-v2.4`) but with the outlines of /dollar reverted to those from v2.003 (e6bc8a20a00c17c7853fbf50581ef9c2154e8187).


MariannaPaszkowska:
@Hoolean, I think we should revert the import where we merged Octavio's liga work, so we have the least diffs in the 2.004 release.

Hoolean:
Sounds good - are there any source changes on `main` since the v2.003 release that we need to actively roll back, or do we just need to close this PR? Looking [at what has landed](https://github.com/googlefonts/googlesans-flex/compare/v2.003...main), we might only have pipeline changes on our `main` branch since then :)

Hoolean:
I've now reviewed the diff between v2.003 and current `main`: there are a few changes, but we have not done a full import since the last release.

### Diffy changes

* Update requirements
     - Makefile
     - requirements.txt
     -  sources/config.yaml

* Increase WinAscender
    - sources/GoogleSansFlex-*.ufo/fontinfo.plist

### Non-diffy changes

* Workflow updates:
    -  .github/
    -  qa/
    -  scripts/

* Syncing **design** sources (not UFOs)
    - sources/design-source/GSF-full.glyphspackage/

Hoolean:
Which if any of the diffy changes do we want to rollback before the workspace fix release?

- Update requirements
  - [x] Do not rollback, as these will be replaced by the version bump we do to get the fonttools fix anyway
- Increase WinAscender
  - [x] ~Consider rolling back?~
      My local tags were outdated, this has already been released and so should potentially stay as-is

PR #1064, opened 2024-10-25
Fix design sources to point at v2.003 release

Hoolean:
We want these to match the Glyphs sources used to build v2.003, although the sources previously added in #1022 for this purpose were a little too new.

After this PR, they should match, and so we will have a correct record and template to branch new design work off of.


Hoolean:
I've added a manual commit to make sure the timestamp matches [what we shipped](https://github.com/googlefonts/googlesans-flex/blob/v2.003/sources/GoogleSansFlex-wg1-wd100-oz144-GD0-RD0-sl-10.ufo/fontinfo.plist#L120-L121); this will minimise unwanted diffs when we use these sources as a template for future work too.

Hoolean:
I've also cancelled the CI jobs, as we do not need to build the UFOs; no changes have been made there, only to the design sources

Hoolean:
(this sort of issue will be a one-off, as Ricky automated the process in #1023 💪)

PR #1068, opened 2024-10-28
Refactor glyphset check to ignore glyph order

### LINKED ISSUE #1067: Ignore glyph order in charset QA check
From discussion in meeting: the order is not important; we should still check for glyphs being added or removed though

Hoolean:
Fixes #1067

### DIFF
--- a/qa/checks/charset.py
+++ b/qa/checks/charset.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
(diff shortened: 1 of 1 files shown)


Hoolean:
Will undraft depending on CI results

Hoolean:
CI results are giving the correct `FAIL`: the check states that the two new glyphs are missing but is not otherwise considering order

Hoolean:
Ready for review and merge

PR #1069, opened 2024-10-28
Import from Octavio-v2.005-add-glyphs for v2.005

Hoolean:
Octavio-v2.005-add-glyphs at 4c3314b6f1


Hoolean:
Looks like this PR isn't picking up on the email I use for this project... I'll close it as I suspect the issue will be resolved if the commit author opens it instead

PR #1072, opened 2024-10-30
Switch to smaller runners for most jobs

### LINKED ISSUE #963: Test smaller GitHub Actions runner performance
Opening this issue following our investigation into #961.  We're using the largest googlefonts org GitHub Actions runners to build and CI test the Flex project.  This might be more metal than we really need and it is the most expensive runner environment, billed on per minute use.  Let's test the memory / CPU requirements to confirm that we need the 256GB + 64 core environment.

Other supported options include:

6gKNHbvZgZpvewe

I also suggest that we throw the higher performance environments at the most computationally intensive tasks.  We could use an approach to throw more computation power at compiles and less at CI tests or vice versa based on actual requirements, rather than a one size fits all approach across everything that we run in the GitHub Actions environment.

Not urgent. But let's look into it during Q3. We want to be able to support fast compiles and fast tests.  There is no intent to slow development times to reduce cost.  Let's attempt to avoid unnecessary costs if they are truly unnecessary and our development time is not impacted by the change.  Thanks.

RickyDaMa:
As requested in #963, moves almost everything to 8 core runners

I've left the diffenator workflow on the big runner as it's known that diffenator can take full advantage of the specs and it takes a fair while to run even doing that

Do not merge this PR unless CI is green - we may hit our timeouts now that we're using weaker runners

### DIFF
--- a/.github/workflows/build-glyphs.yml
+++ b/.github/workflows/build-glyphs.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   convert:
-    runs-on: googlefonts-64cores-256GB
+    runs-on: googlefonts-8cores-32GB
     timeout-minutes: 20
     steps:
     - name: Checkout main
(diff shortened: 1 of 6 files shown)


RickyDaMa:
Looks like we didn't hit the timeouts, FB fails instead making CI red. This PR should be fine to merge

Issue #1070, opened 2024-10-28
Add Font Bakery check to ensure that creation date in `head` does not change between releases


Hoolean:
This has seen a recent fix in the sources, but it would be good to have a small automated check to verify that it has propagated through all active design branches

PR #1084, opened 2024-11-08
Check that creation date in `head` is maintained across releases

Hoolean:
Implements #1070

Ocassionally we have had to reset this field manually:

- 03e557a4eea3dfa6564a8b97a9d90cc509a19655
- 9806e34acb8c881c2aa94dade1b72f0373ffdb85

and while the root issue should now be addressed (corrected in Glyphs sources on main), automated tests can lend full confidence that this metadata will stay the same forever.


Hoolean:
(the `modified` date field does not need checking as it assumes a new timestamp on each build and release)

PR #1109, opened 2024-12-12
ROND avar1

### LINKED ISSUE #1107: Add avar1 table for ROND axis mapping
THe mapping we need is:
ROND 70 => ROND 75

RickyDaMa:
Fixes #1107 

Also updates the import script so this will won't be overriden by a fresh import


RickyDaMa:
We don't particularly like the idea, but we should check with Dave Crossland before closing this

MariannaPaszkowska:
Ok to close, thanks for working on it.

PR #1119, opened 2024-12-30
METADATA.pb subset script & test

### LINKED ISSUE #1092: Add automation to update the subset definitions in metadata.pb
We need to update our subset definitions, and ensure that we use a set of them that include every encoding present in the font—otherwise, missing encodings will be subset out. This is automated on the public families. This would help keeping the subset definitions in sync between the internal METADATA.pb (CSS API pipeline files) files and the ones in the repository. (thanks for advice here, @chrissimpkins )

@simoncozens mentioned that it requires a slight tweak to gftools-add-font, and then simply running it in the GSF directory. That should produce a METADATA.pb file with all the subsets.
Here is a PR prepared by Simon: https://github.com/googlefonts/gftools/pull/1065 that allows run gftools-add-font on a repo and get a GF-like METADATA.pb without being in a known libre license subdirectory.

- [ ]  @m4rc1e Could you have a look at how we can integrate this automated approach of defining that metadata file here?  

- [x] We also should consider promoting the above Font Bakery check from a WARN to a FAIL? (thanks for suggestion @Hoolean )

- [ ] @chrissimpkins  Below are the needed subsets I identified, but we should update it once we have an automated workflow for that.

```
latin, 
latin-ext 
vietnamese 
menu
math 
symbols 
```

RickyDaMa:
Adds @m4rc1e's [Gist script](https://gist.github.com/m4rc1e/8fe31c10352a23cb8974cf25d1bcdca0) to the repo and adds a Fontbakery check to compare the current METADATA.pb with the script's output. If they differ, it's a Fontbakery fail

### DIFF
--- a/qa/check-googlesans.py
+++ b/qa/check-googlesans.py
@@ -36,6 +36,7 @@ PROFILE = {
             "googlesansflex/vf/fvardefault",
             "googlesansflex/opentype/global_fu_attributes",
             "googlesansflex/android_ymin_ymax",
+            "googlesansflex/metadata_subsets",
         ]
     },
     "exclude_checks": [
--- a/qa/checks/googlesans.py
+++ b/qa/checks/googlesans.py
@@ -415,3 +415,5 @@ def com_google_fonts_check_metadata_subsets(font: Font, ttFont: TTFont):
                 ),
             ),
         )
+    else:
+        yield PASS, "subsets available in font and specified in METADATA.pb match"
--- a/qa/checks/googlesans.py
+++ b/qa/checks/googlesans.py
@@ -387,7 +387,7 @@ def com_google_fonts_check_metadata_subsets(font: Font, ttFont: TTFont):
         allow_unknown_field=True,
     )
 
-    expected_raw = subprocess.check_output(
+    suggested_raw = subprocess.check_output(
         (
             "venv_bakery/bin/python",
             "scripts/google_fonts_metadata_subsets.py",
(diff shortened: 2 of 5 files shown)


qa/check-googlesans.py
```diff
@@ -35,6 +35,8 @@
             "com.google.fonts/check/googlesansflex/vf/axis_names",
             "com.google.fonts/check/googlesansflex/vf/fvardefault",
             "com.google.fonts/check/googlesansflex/opentype/global_fu_attributes",
+            "com.google.fonts/check/googlesansflex/android_ymin_ymax",
```
Hoolean: Thanks for adding this 🤜🤛

qa/checks/googlesans.py
```diff
@@ -27,6 +29,8 @@
     compute_unicoderange_bits,
     unicoderange_bit_name,
 )
+from fontTools.ttLib import TTFont
+from google.protobuf import text_format
```
Hoolean, line 33: If we are depending on this now we should add it to our top-level QA requirements

scripts/google_fonts_metadata_subsets.py
```diff
@@ -0,0 +1,69 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import unicodedata as uni
+from argparse import ArgumentParser
+
+from fontTools.ttLib import TTFont
+from gfsubsets import CodepointsInSubset, ListSubsets
```
Hoolean, line 21: Same for this requirement too

Hoolean (commented):
LGTM, just a few indirect dependencies to eliminate before merge

Hoolean:
Thanks for the script @m4rc1e :)

One point I want to check: the code prefers splits where there is a big initial subset, over splits that use a lesser number of subsets, e.g.

Biggest initial subset (current output):
```
A A A A A B C D
```

...versus fewest total subsets:
```
E E E E F F F F
```

As we begin to deploy the script more widely and enforce its output by check, would we prefer the latter, or does the former give something more semantic for our purposes? (apologies if this was already discussed and gone over, I'm joining the subset discussion late in the game)

belluzj:
I've rebased and fixed conflicts. The script now fails here: 
image

We'll investigate more tomorrow

MariannaPaszkowska:
note to self: check the metadata.pb output and the subset definition

MariannaPaszkowska:
@belluzj this check should be a warn in our case. 
For example, we are required to add the `menu` subset.

belluzj:
Note: instead of reviving this, we could look into moving it to fontspector. Also the output of the script is not really trusted by Marianna, because it gives unreliable results sometimes, and we don't know why yet.

PR #1121, opened 2025-01-09
DO NOT MERGE THIS! It is just an experiment to see how the repo would deal with the latest fontbakery from git

felipesanches:
(in preparation to actually cut a stable release if the experiment gives good results)

### DIFF
--- a/requirements-fb.txt
+++ b/requirements-fb.txt
@@ -4,5 +4,5 @@
 # To unpin, delete requirements-fb.txt and get rid of any extra constraints
 # added here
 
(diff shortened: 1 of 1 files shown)


felipesanches:
The only issue that I noticed here was addressed at https://github.com/fonttools/fontbakery/commit/5dec9b121bf6b8d42cf271aaa31b7f23d5b09c75 and now FontBakery v0.13.0 was released.

Please let me know if you have any issue with the new release and I'll be as quick as possible to fix any problems that may surface.

PR #1135, opened 2025-03-04
Add GitHub workflow permissions necessary for release

RickyDaMa:
Also adds some logic to automatically install cairo if not found (as this repo doesn't use f-actions/font-bakery)


Hoolean:
LGTM - I'm going to prune 14b9a10b3c7996b31513fbf68db5b986b6c9cc45 for now to avoid shenanigans on macOS (and my WSL 😁), but we should consider restoring it or something similar at some point for when we test or if we move to plain Ubuntu runners.

Hoolean (approved):

PR #1136, opened 2025-03-04
Install cairo when the CI does not already include it

Hoolean:
Partitioned from @RickyDaMa's PR at #1135:

We use the paid Google Fonts runners on this repository, so this is not normally an issue, but is invaluable for when we test with (or if we move in the future to) plain Ubuntu runners.


Hoolean:
Will tweak a little so it doesn't try to run locally

PR #1150, opened 2025-06-10
Add dependabot configuration for GH Actions version update monitoring

chrissimpkins:
Adds a dependabot configuration to monitor and propose PR updates of our GitHub Actions dependency versions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  # GitHub Actions dependencies
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "00:00"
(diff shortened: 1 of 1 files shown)


chrissimpkins:
CI is failing on https://github.com/googlefonts/googlesans-flex/issues/1151

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  # GitHub Actions dependencies
+  - package-ecosystem: "github-actions"
+    directory: "/"
```
Hoolean: New finding after integrating this with Mono: we'll need to specify our workflow's child actions explicitly to ensure their versions are checked (and so bumped in the resulting PRs) too: googlefonts/googlesans-mono#107

PR #1157, opened 2025-07-01
Add draft script for producing calculator variant

Hoolean:
This won't be maintained, built, or released here, but is a good place to collaborate on the source until it can ship.

## TODO

- [x] Confirm and fill final codepoints to swap
  - Surfaced as configurable with global constant instead
- [x] Confirm and fill final codepoints to subset to
  - Surfaced as configurable with global constant instead
- [x] Test before/after font for unwanted changes


MariannaPaszkowska:
Thank you @Hoolean !

madig:
@MariannaPaszkowska I updated the script with your list

PR #1184, opened 2025-08-26
Use uv for venv management

madig:
Note that there could be one more simplification, namely using a pyproject.toml and a .python-version with a uv.lock file. This would allow us to drop various arguments to uv in the Makefile. Maybe best done when all dependencies are updated, so as to not change too much now.


madig:
To be done when I pick it back up: The testing workflow is imported by the building workflow, but will take the file from the main branch, so the necessary changes in this branch aren't picked up. Need to test in a repo where I can push to main directly or at least change the branch that is imported from.

PR #1191, opened 2025-09-04
Use fontspector where possible

madig:
This seems quite straightforward. Given that fontspector seems to try and use the same check IDs, we can run the checks covered by fontspector in fontspector and the rest in fontbakery.


madig:
More to do, when porting checks: Look into using https://chat.google.com/room/AAAAgistbQI/8R0xfItFDa4/G10DyZE6vQk?cls=10

.github/workflows/release.yml
```diff
@@ -116,6 +116,10 @@ jobs:
       uses: f-actions/opentype-sanitizer@v3
       with:
         path: fonts/android/*.ttf
+    - name: Install cargo-binstall
+      uses: cargo-bins/cargo-binstall@v1.15.3
+    - name: Use cargo-binstall to grab precompiled fontspector
+      run: cargo binstall fontspector@1.3.0
```
Hoolean: We could use the [setup-fontspector](https://github.com/fonttools/setup-fontspector) tool here to have a better match with [the template](https://github.com/googlefonts/googlefonts-project-template). This would make it easier to opt into custom fontspector features too (e.g. Python support) if we need it in the future too

Makefile
```diff
@@ -50,7 +50,7 @@ workspace: build.stamp
 		fonts/workspace
 	mkdir -p fonts/tv
 	mv fonts/workspace/GoogleSansFlexTV[wght].ttf fonts/tv
-	$(UV_RUN) pyftsubset fonts/tv/GoogleSansFlexTV[wght].ttf --output-file=fonts/tv/GoogleSansFlexTV[wght].ttf --unicodes="U+D-25CC,U+FB00-1D61E" --layout-features="tnum,numr,subs,sups,frac,ordn,dnom,zero,kern,locl,mark,mkmk,ccmp,liga" --recalc-average-width --recalc-max-context --recalc-bounds --notdef-outline
+	$(UV_RUN) pyftsubset fonts/tv/GoogleSansFlexTV[wght].ttf --output-file=fonts/tv/GoogleSansFlexTV[wght].ttf --unicodes="U+D-25CC,U+FB00-1D61E" --layout-features="tnum,numr,subs,sups,frac,ordn,dnom,zero,kern,locl,mark,mkmk,ccmp,liga" --name-IDs="*" --recalc-average-width --recalc-max-context --recalc-bounds --notdef-outline
```
Hoolean, line 53: Is this leftover from testing, or something that fontspector highlighted that we're taking the time to fix here?

qa/check-googlesans.toml
```diff
@@ -0,0 +1,32 @@
...
+    "googlefonts/STAT/axisregistry",                   # https://github.com/fonttools/fontbakery/discussions/4214
+    "googlefonts/varfont/generate_static",             # We cut our instances from the VF and will know if that fails
+    "googlefonts/varfont/has_HVAR",                    # The Android font has no HVAR, so we had to modify this check into `googlesansflex/varfont/has_HVAR`
+    "name/family_and_style_max_length",                # we know our statics exceed this limit and it's okay
+    "opentype/family/bold_italic_unique_for_nameid1",  # Expected and desired
+    "opentype/family/consistent_family_name",          # intended with our statics
+    "opentype/varfont/family_axis_ranges",             # our workspace fonts intentionally change axes ranges
+    "opentype/varfont/regular_opsz_coord",             # No, opsz=18
+    "opentype/varfont/valid_default_instance_nameids", # Bogus
+    "outline_alignment_miss",                          # Separate.
```
Hoolean:
```suggestion
    "outline_alignment_miss",                          # Separate, covered by neighbouring check-outlines profile.
```

qa/check-googlesans.toml
```diff
@@ -0,0 +1,32 @@
...
+    "contour_count",                                   # design rather than QA problem
+    "dsig",
+    "family/single_directory",                         # conflicts with gftools' folder structure
+    "family/vertical_metrics",                         # GS is our reference.
+    "file_size",                                       # We're going bigger
+    "fontdata_namecheck",                              # online resource unavailable https://github.com/fonttools/fontbakery/issues/2719
+    "googlefonts/font_names",                          # We have our own naming ideas
+    "googlefonts/glyph_coverage",                      # We have our own target
+    "googlefonts/glyphsets/shape_languages",           # we do our own shaperglot check
+    "googlefonts/name/family_name_compliance",         # Irrelevant for us, but see https://github.com/fonttools/fontspector/issues/414.
```
Hoolean, line 13: Out of interest, were we failing this? 👀

qa/check-googlesans.py
```diff
@@ -12,10 +12,31 @@
...
 from fontbakery.profiles.googlefonts import PROFILE as GOOGLEFONTS_PROFILE
 
+# Try and exclude all checks that are covered by fontspector.
+try:
+    data_json = subprocess.check_output(
+        ["fontspector", "--profile", "googlefonts", "--list-checks-json"]
+    )
+    data = json.loads(data_json)
+    fontspector_checks = [check["id"] for _, checks in data.items() for check in checks]
+    print("INFO: Skipping checks covered by fontspector.")
```
Hoolean: This does look fairly robust, although is it only the custom checks that survive this? We could always put those into a separate profile to avoid the complexity of the hybrid

scripts/fontbakery.sh
```diff
@@ -68,16 +68,24 @@ fi
 
 # Compiled font tests
 echo "$all_ttfs" \
-    | xargs $FONTBAKERY check-profile -l WARN --auto-jobs --succinct --no-progress \
-    --html out/fontbakery/fontbakery-outlines-report.html \
-    qa/check-outlines.py {} \
-    || failed+=("fontbakery.profiles.outline")
+    | xargs fontspector --profile googlefonts --configuration qa/check-outlines.toml \
```
Hoolean, line 71: Do we still need to specify the profile if we list all of the checks explicitly?

Hoolean (commented):
LGTM, tho a few Qs on particular

FYI @simoncozens for Flex in particular, we might use fontspector in pure-Rust mode and invoke Font Bakery in Python separately for the other custom checks, as we already have them split into easily separable profiles. As each gets ported, let's toggle the profiles to Fontspector one-by-one 🚀

Hoolean:
@madig the test CI is failing:

```
xargs: fontspector: No such file or directory
```

Makefile
madig: The latter. pyftsubset is dropping name IDs that fontbakery/-spector look for (specifically, ID 9, Designer).

qa/check-googlesans.toml
madig: Yes. For some reason fontbakery wasn't complaining, but fontspector is.

scripts/fontbakery.sh
madig: We're listing _exclusions_, so you need to specify the profile _and_ the configuration, but you seemingly can't do it in one go if I understand Simon correctly.

madig:
It is failing here because the test Action is pulling the workflow from main. It works in our test repo.

qa/check-googlesans.py
madig: Hm, we could. My idea was that we'd port the checks piecemeal anyway and can just delete them here. As long as the checks have the same ID, this should be robust. There are some fontbakery checks that haven't made it to fontspector (yet?) so these are caught as well.

MariannaPaszkowska:
@madig could you please give us the list of checks that are included here and which ones still need to be ported?

madig:
Speaking just about the "googlefonts" profile, fontspector is "missing" these checks:

- caps_vertically_centered
- designspace_has_consistent_codepoints
- designspace_has_consistent_glyphset
- designspace_has_consistent_groups
- designspace_has_default_master
- designspace_has_sources
- family/single_directory
- fontbakery_version
- googlefonts/colorfont_tables
- googlefonts/description/family_update
- googlefonts/family_name_compliance
- googlefonts/metadata/axisregistry_bounds
- googlefonts/metadata/axisregistry_valid_tags
- googlefonts/metadata/canonical_style_names
- googlefonts/metadata/canonical_weight_value
- googlefonts/metadata/category_hints
- googlefonts/metadata/consistent_axis_enumeration
- googlefonts/metadata/date_added
- googlefonts/metadata/designer_values
- googlefonts/metadata/empty_designer
- googlefonts/metadata/filenames
- googlefonts/metadata/includes_production_subsets
- googlefonts/metadata/match_filename_postscript
- googlefonts/metadata/match_fullname_postscript
- googlefonts/metadata/match_name_familyname
- googlefonts/metadata/match_weight_postscript
- googlefonts/metadata/menu_and_latin
- googlefonts/metadata/minisite_url
- googlefonts/metadata/nameid/family_and_full_names
- googlefonts/metadata/nameid/font_name
- googlefonts/metadata/nameid/post_script_name
- googlefonts/metadata/parses
- googlefonts/metadata/single_cjk_subset
- googlefonts/metadata/subsets_order
- googlefonts/metadata/undeclared_fonts
- googlefonts/metadata/unique_full_name_values
- googlefonts/metadata/unique_weight_style_pairs
- googlefonts/metadata/unsupported_subsets
- googlefonts/metadata/valid_filename_values
- googlefonts/metadata/valid_full_name_values
- googlefonts/metadata/valid_post_script_name_values
- googlefonts/production_glyphs_similarity
- googlefonts/repo/fb_report
- googlefonts/repo/sample_image
- googlefonts/repo/upstream_yaml_has_required_fields
- googlefonts/repo/zip_files
- googlefonts/varfont/generate_static
- opentype/cff2_call_depth
- opentype/cff_ascii_strings
- opentype/cff_call_depth
- opentype/cff_deprecated_operators
- opentype/dsig
- opentype/gdef_mark_chars
- opentype/gdef_non_mark_chars
- opentype/gdef_spacing_marks
- opentype/kern_table
- opentype/unwanted_aat_tables
- ots
- shaping/collides
- superfamily/list
- superfamily/vertical_metrics
- ttx_roundtrip
- ufo_features_default_languagesystem
- ufo_recommended_fields
- ufo_required_fields
- ufo_unnecessary_fields
- ufolint
- varfont/bold_wght_coord
- varfont/duplicate_instance_names

@simoncozens has said that these should be redundant checks that have been rolled into different ones in fontspector, or source checks that he is not planning on supporting currently. I see there's also the "ots" check, which I suppose should be kept around.

If Simon can confirm that this list is as expected and benign, I can drop the fontbakery non-source and non-ots checks.

As for the other custom checks we have, they remain untouched until I get around to porting them (or in the case of source checks, just keep them in fontbakery):

- googlesans/features/regression
- googlesans/features/variableuprights
- googlesans/glyphs/glyphset-contents
- googlesansflex/android_hvar
- googlesansflex/android_ymin_ymax
- googlesansflex/opentype/global_fu_attributes
- googlesansflex/opentype/head/created
- googlesansflex/opentype/os2/unicode_range_bits
- googlesansflex/sources/decomposed_by_mix
- googlesansflex/sources/decomposed_by_skip
- googlesansflex/sources/decomposed_by_var_transform
- googlesansflex/sources/kerning_present
- googlesansflex/sources/same_tabular_width
- googlesansflex/sources/suspicious_kerning_values
- googlesansflex/varfont/has_HVAR
- googlesansflex/vf/axis_names
- googlesansflex/vf/fvaraxes
- googlesansflex/vf/fvardefault

madig:
I looked at the stock fontbakery `ufo*` checks and I don't think they have ever been run on GSF.

1. They only run if you point fontbakery at UFOs, not the Designspace like we do with our custom source checks.
2. Some of them use defcon-isms with ufoLib2 and crash. So, the checks are broken upstream.

Switching to fontspector thus does not change anything for GSF :)

simoncozens:
I can confirm that your list of missing checks is expected - each of them either deliberately disabled / no longer felt to be useful or merged into other checks.

madig:
Fantastic! Thanks :)

PR #1194, opened 2025-09-05
Use fontc

madig:
To do: 

* [x] Regression test
* [x] Diffenator3 test


madig:
I ran diffenator3 on the fontmake and the fontc compiled font and diffenator3 says (ignoring the table diffs we know about already):

image

Meaning that the table diffs seem to not matter to the actual rendering and layouting of the font, as seen by diffenator3.

To be surer, I'll run it on all master locations and maybe on the locations inbetween them as well. Let's see.

Hoolean:
Let's see if we can pin down the exact cause of one of the `gvar` diffs, too

madig:
I ran diffenator3 and had it look at a large amount of locations inbetween masters (`--cross-product --cross-product-splits 2`), and it found no diffs.

Hoolean:
We manually diff'd a build from the older fonttools version (which we pin against now) with the newest build from fontc. We have largely ruled out any semantic differences on the ttx level.

This includes:

- `gvar`
   - Some noop deltas have been removed
- `GPOS` and `GDEF`
   - _otl-normaliser_ still shows identical results

The last areas we'd want to check to have decent confidence:
- `HVAR`
   - We still see some differences here; we are discussing with @anthrotype, as I think we'd looked into this before
- `GSUB`
   - There are minor differences in lookup arrangements/construction; we'll think of a way to assess these

madig:
I have manually tested the standout GSUB changes (fontc puts the L-periodcentered-L and l-periodcentered-l subs and the I-J-IJ and i-j-ij subs into their respective own lookup, while old fontmake seems to split them up into 4 separate ones?) and the fontc-made font works identically.

If we can explain away the HVAR diffs, we can merge :)

Hoolean:
Good news for `HVAR`:
- The latest version of fonttools matches the latest version of fontc
- The diff from old to new fonttools was originally motivated by the Flex diff against fontc (https://github.com/fonttools/fonttools/commit/548ffda37bf6a36c4bd3f45255d178d371bb9cf9)

Because of these, and the positive diffenator3 results above, we have confidence that the `HVAR` changes are harmless.

madig:
Harry and me have updated the internal "GS Code & Flex fontmake vs fontc Builds Comparison" doc with a summary of our testing approach and the results.

MariannaPaszkowska:
I completed all tests (diffenator 2 and 3, ttx comparison, functionality, etc.) and I didn't find any regressions at this point. 

To do: Check for byte level differences vs TTX level differences (like issues within a given font table, something that will be sanitized when ttx dumping), to avoid issues like the recent one we had in ([ GS Code, link to the issue here](https://github.com/googlefonts/googlesans-code/issues/23))

PR #1197, opened 2025-09-29
Add BASE table to feature code (produced by `autobase`)

### LINKED ISSUE #1170: Add `BASE` table
More details to follow...

RickyDaMa:
- Update requirements.txt so `fontTools` supports the BASE table
- Add autobase config file and plumb into Makefile
- Add fea code produced by autobase (edited, as currently it's not completely valid/complete, but the important metrics are there)
- Stop the subsetter from dropping the BASE table (upstream solution also being considered: https://github.com/fonttools/fonttools/pull/1866)

TODO:
- [x] Add to Glyphs sources
- [x] CI check for presence of BASE

### DIFF
--- a/sources/GoogleSansFlex-wg1-wd100-oz144-GD0-RD0-sl-10.ufo/features.fea
+++ b/sources/GoogleSansFlex-wg1-wd100-oz144-GD0-RD0-sl-10.ufo/features.fea
@@ -71,9 +71,7 @@ table BASE {
     HorizAxis.BaseScriptList
         grek romn 0,
         latn romn 0;
-    HorizAxis.MinMax grek dflt -466, 1110;
     HorizAxis.MinMax latn dflt -778, 2699;
-    HorizAxis.MinMax latn VIT  -618, 2391;
 } BASE;
 
 
--- a/sources/GoogleSansFlex-wg1-wd100-oz144-GD0-RD0-sl-10.ufo/features.fea
+++ b/sources/GoogleSansFlex-wg1-wd100-oz144-GD0-RD0-sl-10.ufo/features.fea
@@ -364,3 +364,13 @@ sub t f by t_f;
 sub t t by t_t;
 
 } dlig;
+
+table BASE {
+    HorizAxis.BaseTagList romn;
+    HorizAxis.BaseScriptList
+        grek romn 0,
+        latn romn 0;
+    HorizAxis.MinMax grek dflt -466, 1110;
+    HorizAxis.MinMax latn dflt -778, 2699;
+    HorizAxis.MinMax latn VIT  -618, 2391;
+} BASE;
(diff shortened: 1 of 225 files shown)


sources/GoogleSansFlex-wg1-wd100-oz144-GD0-RD0-sl-10.ufo/features.fea
```diff
@@ -65,6 +65,18 @@ lookup ccmp_soft_dotted {
...
 
+# Prefix: BASE
+table BASE {
+    HorizAxis.BaseTagList romn;
+    HorizAxis.BaseScriptList
+        grek romn 0,
+        latn romn 0;
+    HorizAxis.MinMax grek dflt -466, 1110;
+    HorizAxis.MinMax latn dflt -778, 2699;
+    HorizAxis.MinMax latn VIT  -618, 2391;
```
Hoolean: Something we need to investigate: based on our preliminary investigations we would expect Vietnamese to need more generous bounds, but at the moment they are more restrictive

RickyDaMa:
Blocked on https://github.com/googlefonts/fontc/pull/1675

sources/GoogleSansFlex-wg1-wd100-oz144-GD0-RD0-sl-10.ufo/features.fea
simoncozens: Note that the Latin wordlists contain the letter Ǘ, which is very tall.

RickyDaMa:
Despite updating autobase to the latest version (the same used for Google Sans), the metrics haven't changed, just now the Greek & Vietnamese entries have been dropped as the Latin metrics eclipse them both

Here's the fontheight report and the TTF used to produce it (built from this branch): [Google Sans Flex BASE report.zip](https://github.com/user-attachments/files/23729428/Google.Sans.Flex.BASE.report.zip)

simoncozens:
Something's wrong with these as the libreoffice and aosp lists don't seem to be segmented by script - all the words are Latin or Common script word.

RickyDaMa:
@simoncozens I'm not sure I see what you mean. The reason why there's very few words under other scripts is because Flex can't shape them. If there are any `.notdef`s in the glyph stream, it's excluded ([link](https://github.com/googlefonts/fontheight/blob/354a9539dc0aee6ff03f04877f2143e0090ebf3f/core/src/lib.rs#L298-L306)). As for AOSP Malay/Tagalog, they don't have script tags so get lumped into the "Unknown" section

RickyDaMa:
Hi @MariannaPaszkowska, I've updated the BASE table again with the exclusion for "Ǻ". We're now getting a seperate entry for Vietnamese as we hoped/expected

See the attached report, you can now see that Ǻ would clip based on BASE metrics: 
[GSF BASE exclusions.zip](https://github.com/user-attachments/files/24100490/GSF.BASE.exclusions.zip)

PR #1199, opened 2025-09-30
For b/422910732, I'm using this commit to tag a v2.007 release

davelab6:
I will attach just the Android TTF file, made by hotfixing the v2.006 Android TTF as follows, to the release page tied to this commit

```sh
ttx -x HVAR GoogleSansFlex\[GRAD\,ROND\,opsz\,slnt\,wdth\,wght\].ttf -o temp.ttx
vi temp.ttx # edit v2.006 to v2.007
ttx temp.ttx
mv temp.ttf GoogleSansFlex\[GRAD\,ROND\,opsz\,slnt\,wdth\,wght\].ttf
```

(Also I fixed a few CHANGELOG typos)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,18 +1,22 @@
 # Google Sans Flex Changelog
 
-## Version 3.006 (2028-08=25)
+## Version 2.007 (2025-09-30)
+## Changed
+- removed HVAR table by hotfixing the Android v2.006 TTF 
+
+## Version 3.006 (2025-08-25)
 ### Changed
 - added currency signs: ₹ ₩ ฿ (#1164)
 - added script to find where glyphs accidentally vary width across duplex axes ([#1180](https://github.com/googlefonts/googlesans-flex/pull/1180))
 
 
-## Version 3.005 (2028-08-19)
+## Version 3.005 (2025-08-19)
 ### Changed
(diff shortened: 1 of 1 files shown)


davelab6:
Here's the ZIP to attach

[GoogleSansFlex-v2.007.Android.zip](https://github.com/user-attachments/files/22625239/GoogleSansFlex-v2.007.Android.zip)

Hoolean:
@davelab6 Morning Dave!

Unfortunately, v2.006 was cut before https://github.com/fonttools/fonttools/pull/3912 was addressed; this means that removing its `HVAR` triggers incorrect widths on many styles and glyphs in Android.

To avoid this, we'd need to rebuild with `USE_MY_METRICS` unset. We could do this by branching from the v2.006 tag, but building with [an additional command line arg](https://github.com/googlefonts/googlesans-flex/blob/c1a4a35a20e8e74a99bd54d6a179604163151d7f/sources/config.yaml#L20-L21). Would you like us to prepare that?

davelab6:
@Hoolean perfect, yes please. This isn't extremely urgent but if you can do it pretty soon that will be great

Hoolean:
Smashing -- we'll put together a `clean-2.006` branch shortly 😄

PR #992, opened 2024-07-22
Add WIP script to report on the kerning issue in v2.003

belluzj:
Adds scripts used for filtering kerning for v2.003 release and subsequent work on the sources.

I used this command to scrub the (massive) copies of glyphs sources that were in the branch:

```
git filter-branch --index-filter 'git rm -rf --cached --ignore-unmatch scripts/one-off/diff-kerning/GSF-full-2.002.glyphs scripts/one-off/diff-kerning/GSF-full-2.003.glyphs scripts/one-off/diff-kerning/GSF-full-2.003.glyphspackage scripts/one-off/diff-kerning/GSF-full-2.002.glyphspackage' origin/main..HEAD
```

PR #1219, opened 2025-11-10
Clean kerning pairs for undefined groups

belluzj:
Leftovers from https://github.com/googlefonts/googlesans-flex/pull/992

@simoncozens FYI


MariannaPaszkowska:
Thank you, @belluzj ! We will be importing soon, so just noting here that we will hae to run this script again on the sources after we import.

PR #1221, opened 2025-11-21
Update issue templates

MariannaPaszkowska:


RickyDaMa:
I've rebased this branch so that it includes #1222, which should now allow the pipeline to pass

PR #1271, opened 2026-02-18
Build derived font for Figma with 'Variable' suffix

Hoolean:
This is identical to the release font, other than family name and its derived name records.

All of the current variable font release's instances are at the default `opsz` position, and so no further modification is required to accommodate Figma's auto-optical sizing behaviour.

## TODOs

- [x] Test new workflow
- [ ] `fvar` instance postscript names are not renamed; do they need to be?


Makefile
```diff
@@ -44,6 +44,15 @@ android: build.stamp
...
 		fonts/android/GoogleSansFlex[GRAD,ROND,opsz,slnt,wdth,wght].ttf
 
+figma: build.stamp
+	mkdir -p fonts/figma
+	-@rm fonts/figma/*.ttf
+	$(UV_RUN) gftools-rename-font \
+		fonts/variable/GoogleSansFlex[GRAD,ROND,opsz,slnt,wdth,wght].ttf \
+		--suffix \
+		" Variable" \
+		-o fonts/figma/GoogleSansFlex[GRAD,ROND,opsz,slnt,wdth,wght].ttf
```
RickyDaMa: Did we ought to make the file name `GoogleSansFlexVariable[GRAD,ROND,opsz,slnt,wdth,wght].ttf` to match?

Hoolean: Good Q! @MariannaPaszkowska Should the "Variable"-suffixed fonts for Figma have an adjusted file name too, or just the original one?

RickyDaMa: image

Fontspector thinks we should

MariannaPaszkowska: Yes, we should update the filenames too

PR #1280, opened 2026-03-03
Remove hardcoded Axis Mapping parameter

RickyDaMa:
Addresses Octavio's issue, raised [here](https://chat.google.com/room/AAAAgistbQI/fKa0dMXlXDU/fKa0dMXlXDU?cls=10)

This lets glyphsLib infer the axes from masters instead of telling it explicitly


MariannaPaszkowska:
Hi all, we can allow this change temporarily and once we are set in the new DS we will go back to hardcoding values. Otherwise things can out of hand pretty quickly. We need to make sure we are not regressing as we import.  And we can think about other approaches.

PR #1282, opened 2026-03-24
`CONTRIBUTING.md`: discuss adding glyphs and shaping changes

RickyDaMa:
Identical to https://github.com/googlefonts/googlesans/pull/727 currently

Do I need to discuss the import process, or are we planning on doing away with that?

Preview [here](https://github.com/googlefonts/googlesans-flex/blob/contributing/CONTRIBUTING.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,37 @@
 Want to contribute? Great! First, read this page (including the small print
 at the end).
 
-### Before you contribute
+# Font source changes
+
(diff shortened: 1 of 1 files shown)


RickyDaMa:
Putting this back in draft as there's some useful content here, but it needs significant adjustments now that Glyphs are our only source format.

TODO: this guide should also include what needs to change if axes are being added/removed.

RickyDaMa:
I've merged this PR into #1307, including adding extra detail about adding/removing axes

PR #1288, opened 2026-05-26
Use Fontspector to run Python TTF checks

### LINKED ISSUE #1301: Downgrade relevant checks to warnings instead of excluding them
Note: Later, we should continue working on downgrading those to warnings instead of ignoring them entirely.

_Originally posted by @MariannaPaszkowska in https://github.com/googlefonts/googlesans-flex/issues/1300#issuecomment-4844821158_

RickyDaMa:
This PR moves all of the binary font checks to Fontspector, including running our custom Python ones using it and the new `fontspectorapi`.

Unfortunately fontspector does not have good support for source checks, so moving ours to it isn't really feasible at this time.

~~This PR is blocked on the release of `fontspectorapi` on PyPI (@simoncozens 🙏)~~ Thanks!


MariannaPaszkowska:
@simoncozens Are you planning to add support for source checks to Fontspector?

simoncozens:
What do you need? 🙂 Do you want actual source checks, or just support for it? Because it's easy enough already to declare your own file type and route your checks to that. I can provide examples if needed.

RickyDaMa:
There are a few upstream checks we use:

- `designspace_has_consistent_codepoints`
- `designspace_has_consistent_glyphset`
- `designspace_has_consistent_groups`
- `designspace_has_default_master`
- `designspace_has_sources`

However given we're imminently looking at moving away from DS+UFO production sources in #1291, an example on how we could set up Glyphspackage as a file type so we can run source checks on that would probably be most relevant, thanks 🙏

MariannaPaszkowska:
@simoncozens just following up on this, could you share some examples with us?

simoncozens:
This is on my list for this week but I am struggling with a few things I need to get done.

PR #1291, opened 2026-06-17
Glyphspackage as production sources (remove DS+UFOs)

RickyDaMa:
Depends on #1288 being merged first (for simplicity's sake).

Diffs:

- gvar is 'only' 99.997% identical compared to when compiling from DS+UFOs. No shaping/pixel regressions though

Changes:

- Removed import workflow (including its components, such as `create-import-branch` and `glyphspackage2ufo`)
- Remove "Build from Glyphspackage" workflow - no longer any different to the regular build workflow
- Removed burndown chart workflow - wasn't being used
- Simplified tests workflow due to removal of "Build from Glyphspackage"
- ⚠️ Comment out Fontbakery source checks
- Update `set-overlap-bits.py` & `glyphs-with-overlap.txt` to work on Postscript names, so reading the sources is no longer needed for this step

TODO:
- [x] Simplify CI
- [x] Test CI
- [x] Update README
- [ ] Port source checks to Fontspector


MariannaPaszkowska:
Do we know how meaningful those 'only' 00.003% of gvar changes are?

MariannaPaszkowska:
From the meeting today: 

_@madig was unable to determine the source of the gvar diffs after 1h. After identifying the glyphs where the diffs were, subsetting the font down to only those glyphs then resulted in a font that compiled with no diffs._

Glyphs in question: ` /quotesingle /quotedbl /colon /period /horizontalelipsis `

RickyDaMa:
GitHub workflows aren't running here for whatever reason (probably because I've made such significant changes compared to `main` that GitHub is a bit scared & confused). I've tested them though and they all work!

Main outstanding question is what sources do we want to commit to `main`, the design sources used to produce the current UFOs on `main`?

PR #1300, opened 2026-06-30
Exclude `googlefonts/family_name_compliance`

RickyDaMa:
Will resolve the failing release pipeline that currently doesn't like some of the workspace fonts

See: https://github.com/googlefonts/googlesans-flex/actions/runs/28394311121


MariannaPaszkowska:
Note: Later, we should continue working on downgrading those to warnings instead of ignoring them entirely.

PR #1307, opened 2026-08-03
Remove Grade axis from builds & flesh out contributor documentation

RickyDaMa:
It's going to remain in the sources, but we no longer want it in the fonts

TODO:
- [x] Builds
- [x] Can we be more robust/flexible about file name changes, in the case of future axis experiments?
- [x] Cut instances
- [x] Tests

[`CONTRIBUTING.md` rendered](https://github.com/googlefonts/googlesans-flex/blob/remove-grade/CONTRIBUTING.md)


CONTRIBUTING.md
```diff
@@ -1,7 +1,54 @@
...
 at the end).
 
-### Before you contribute
+# Font source changes
+
+Beyond the usual changes made using your font editor, there may be some additional changes to make in order for the pipeline to pass.
+
+## Formatting sources
+
+If you want any changes to the Designspace or UFOs, please run `scripts/gs-normalize-designspace.py` to reformat everything appropriately.
```
belluzj: Is this still relevant if the sources are now glyphspackage?

RickyDaMa: Very true, I'll remove that section

qa/check-googlesans.py
```diff
@@ -333,32 +301,25 @@ def fvar_default(font_path: Path) -> CheckStatuses:
...
+            yield FAIL, f"Font has unexpected axis {axis.axisTag}"
+        elif actual != expected:
             yield (
                 FAIL,
                 (
-                    f"Font does not include the correct "
-                    f"fvar {tag} axis default.\n"
+                    "Font does not include the correct "
+                    f"fvar {axis.axisTag} axis default.\n"
                     f"Found: `{axis.defaultValue}` and expected `{expected}`"
```
belluzj:
```suggestion
                    f"Found: `{actual}` and expected `{expected}`"
```

qa/shaping/zero.json
```diff
@@ -355,7 +355,7 @@
         "zeroslash"
       ]
     },
-    "GoogleSansFlexTextFine-Italic[wght].ttf": {
+    "GoogleSansFlexText-Italic[wght].ttf": {
```
belluzj, line 414: A bit weird that the instance entries of the dict are moving around? Should there be a sort() somewhere to make that stable?

belluzj:
LGTM apart from comments above

qa/shaping/zero.json
RickyDaMa: Agreed, but I checked that it disappeared and then re-appeared with the same name. I'll see if it's easy to fix the order

RickyDaMa: Looks like it depends on the order of the Bash glob yields the TTFs currently. I'll add a `sorted` in Python land to iron it out

PR #1308, opened 2026-08-11
GitHub workflows: comply tighter security policies & consolidate files

RickyDaMa:
- Pinned GitHub Actions we used to specific hashes
- Checked that Dependabot will still be able to support updates when we're using hashes - research seems to indicate so
- Inlined `tests.yml` to avoid needing to pin an action defined in our own repository(!)
- Consolidated all build-related workflows & actions into one single linear YAML file(!!)


.github/workflows/build.yaml
```diff
@@ -6,37 +6,360 @@ on:
...
     # on: pull_request uses head_ref for branch name, on: push uses ref_name
     - name: Checkout ${{ inputs.git-ref || github.head_ref || github.ref_name }}
-      uses: actions/checkout@v7
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
       with:
         ref: ${{ inputs.git-ref || github.head_ref || github.ref_name }}
+        persist-credentials: false
     - name: Set up Python
-      uses: actions/setup-python@v6
+      uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
```
Hoolean, line 39: Can uv do this too? One fewer action to rely on; on the other hand, maybe relying on uv for this is less idiomatic

.github/workflows/build.yaml
```diff
@@ -6,37 +6,360 @@ on:
...
+      shell: bash
+      run: |
+        # Prevent font re-building
+        touch build.stamp
+        make workspace
+    - name: Create Android flavour GS Flex
+      run: make android
+    - name: Create Figma flavour GS Flex
+      run: make figma
+    - name: Generate workspace archive name
```
Hoolean, line 107: nit: workspace and TV?

.github/workflows/build.yaml
```diff
@@ -6,37 +6,360 @@ on:
...
+        if-no-files-found: error
+  test-instances:
+    if: ${{ github.event_name == 'release' || inputs.cut-instances }}
     needs:
-    - build
-    uses: googlefonts/googlesans-flex/.github/workflows/tests.yml@main
-    with:
-      branch: ${{ inputs.git-ref || github.head_ref || github.ref_name }}
-      artifact-name: ${{ needs.build.outputs.artifact-name }}
+    - cut-instances
```
Hoolean, line 222: nit: maybe this should be 'derived' instead of instances at this point

.github/workflows/build.yaml
```diff
@@ -6,37 +6,360 @@ on:
...
+      uses: f-actions/opentype-sanitizer@d5603373a9adccd136dd48b037b9b5178b485bea  # v4.0.0
+      with:
+        path: fonts/variable/*.ttf
+    - name: Analyze TTF file size
+      shell: bash
+      run: make font-size
+    - name: Run shaperglot
+      shell: bash
+      run: |
+        touch build.stamp
```
Hoolean, line 197: nit: redundant `touch`?

.github/workflows/build.yaml
```diff
@@ -6,37 +6,360 @@ on:
...
+        make shaperglot
+    - name: Setup Fontspector
+      uses: fonttools/setup-fontspector@d53404291dc78a5a41cf76ea48e1d13a731fb812  # main
```
Hoolean, line 200: Do we need version tags/releases for Dependabot to detect updates here?

.github/workflows/build.yaml
```diff
@@ -6,37 +6,360 @@ on:
...
+
+        ANDROID_ZIP_NAME="GoogleSansFlex-${GITHUB_REF#refs/*/} Android.zip"
+        cd "$GITHUB_WORKSPACE/${ANDROID_ARTIFACT}"
+        # Rename file to match what Android's already using (regardless as to how representative the name is)
+        mv GoogleSansFlex[GRAD,ROND,opsz,slnt,wdth,wght].ttf GoogleSansFlex-Regular.ttf
+        zip -9 "../$ANDROID_ZIP_NAME" *.ttf *.md
+    
+        cd "$GITHUB_WORKSPACE"
+        # Repo must be explicitly stated as we don't have the repo checked out
+        gh release upload --repo $GITHUB_REPOSITORY $GITHUB_REF_NAME \
```
Hoolean, line 360: These must be quoted

qa/googlesans-profile.toml
```diff
@@ -47,6 +47,8 @@ exclude_checks = [
     # google-fonts-axisregistry with this merged:
     # https://github.com/googlefonts/axisregistry/pull/216
     "googlefonts/fvar_instances",
+    # The instances we cut give this check a lot of ammunition
+    "typographic_family_name",
```
Hoolean, line 51: Is this a stray change, or an improvement shipping at the same time? 👁️

Hoolean (commented):
Super, this all looks much simpler 💪 I've tried to scrutinise a little closer than normal, given that it's a larger change, and some bits we'll be applying in a few places.

As it's a workflow, the proof will be in the pudding when running this, although I understand we've tested with success.

.github/workflows/build.yaml
RickyDaMa: Yes, we could let `uv` handle Python installations if we wanted. Not sure what's considered idiomatic

.github/workflows/build.yaml
RickyDaMa: Nope, when pinned to an untagged commit, Dependabot assumes you want the latest on the default branch, which is true here

.github/workflows/build.yaml
RickyDaMa: They're both guaranteed to have no whitespace in them so they don't need to be

qa/googlesans-profile.toml
RickyDaMa: Not sure why it wasn't flagging before, maybe a new check as part of the recent Fontspector release. It's definitely not useful for this repo though, hence excluding it. Every single workspace font makes it annoyed

.github/workflows/build.yaml
RickyDaMa: Yep I think that is, good spot

.github/workflows/build.yaml
RickyDaMa: From https://docs.astral.sh/uv/guides/integration/github/#setting-up-python:

> the official GitHub `setup-python` action can be used. This can be faster, because GitHub caches the Python versions alongside the runner

.github/workflows/build.yaml
Hoolean: I would still add them so that we don't rely on the git branch rules to do our escaping for us, generally prefer it whenever a variable is being used in this way, especially from an external context

PR #1309, opened 2026-08-13
Workflow tidying

RickyDaMa:
Addressing review comments left after #1308 was merged

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -348,7 +348,7 @@ jobs:
     
         cd "$GITHUB_WORKSPACE"
         # Repo must be explicitly stated as we don't have the repo checked out
-        gh release upload --repo $GITHUB_REPOSITORY $GITHUB_REF_NAME \
+        gh release upload --repo "$GITHUB_REPOSITORY" "$GITHUB_REF_NAME" \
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -176,14 +176,7 @@ jobs:
...
     - name: Avoid re-building fonts
       shell: bash
-      run: |
-        if [ -d "fonts/variable" ]; then
-          # Trick make into not re-building the fonts
-          touch build.stamp
-        else
-          echo "::error title=Built fonts not found::No built fonts in ./fonts/variable"
-          exit 1
-        fi
```
Hoolean, line 186: I _think_ we added this because `OT Sanitizer` passes silently if you give it zero arguments

Hoolean (commented):
Ta for speedy PR 🚀 One bit to check, otherwise LGTM

.github/workflows/build.yaml
RickyDaMa: Yeah but now this isn't within a safe-contained action I don't feel the need to validate my inputs. Like, I trust me from elsewhere in the file to make sure the files are actually there.

Hoolean (approved):

PR #1310, opened 2026-08-14
Add Spacing axis with `gftools gen-spac`

### LINKED ISSUE #1312: Add SPAC axis
This is a request from one of the teams to add Spacing axis.

Visually the maximum should mach around +20% tracking.
We need to define the minimum.

The values for min and max should follow what we have in axis registry.
https://github.com/googlefonts/axisregistry/blob/main/Lib/axisregistry/data/spacing.textproto 

Thanks @m4rc1e for suggestion, we can use https://github.com/googlefonts/gftools/blob/main/Lib/gftools/scripts/gen_spac.py

RickyDaMa:
Note: this PR depends on #1307 being merged first

Axis range: -20 to 40

STAT entry added at 0 (elidable name "Default"). No other names planned


MariannaPaszkowska:
Thanks for picking this up, we only want the "Default" for STAT table here.

PR #1323, opened 2026-09-16
Suspicious kerning source check using Fontspector

### LINKED ISSUE #828: Make the WARN Kerning values outside the accepted range of [20, 400] width sensitive
Make the `WARN Kerning values outside the accepted range of [20, 400]` check sensitive to the width axis and calculate new ranges.

### LINKED ISSUE #945: Consider running Fontspector checks also on sources before build
Currently, they're run afterwards, and only if the font builds, which seems illogical

It also means that if the font doesn't build, the sources aren't checked, which may have helped to diagnose the issue

Two approaches we could instead consider:

1. Run the source checks before building the font, and only build if the checks pass (no fails)
    - Should take roughly the same amount of time, maybe slightly longer
    - Might prevent the font from being built when it otherwise would have

2. Run the source checks alongside building the font
    - Will speed up the pipeline, as more work can be done at once
    - We may end up wasting time trying to build a font that we could have known would fail to build

### LINKED ISSUE #1302: Migrate source checks to Fontspector
The second half of #1288, which would allow us to remove our dependency on Fontbakery entirely

Example code upstream: https://github.com/fonttools/fontspector/blob/400028d3c59f2a8d54b3d1e6253ce3e41f083709/python/exampleplugin.py#L68-L82

Needs Fontspector v1.7.2 or later for Glyphspackage support (or more accurately, directory testable support)

We should probably settle on a specific source format (see #1291) before starting on this issue

madig:
This removes fontbakery and ports the suspicious kerning source check to run under fontspector.

Also fixes the WWS warning from fontspector.

Closes #945, #1302, #828


qa/check-sources.py
```diff
@@ -11,29 +13,176 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+#
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "fontspectorapi",
+#     "glyphsLib",
```
RickyDaMa: I'd feel comfier if we pinned the version of glyphsLib we relied on

RickyDaMa (commented):
This looks good, I like the new implementation of the suspicious kerning check - hopefully the sensitivity is at a good level

However, I think we could separate the `test` recipe in the Makefile to have a separate recipe for source checks, and then we could also solve #945 in this PR too. What do you think?

Makefile
```diff
@@ -29,6 +29,12 @@ build: build.stamp
 
 build.stamp: requirements.txt sources/config.yaml $(SOURCES)
 	rm -rf fonts/
+	$(UV_RUN) fontspector \
+		--loglevel warn --succinct --full-lists \
+		--html out/fontspector/source-checks-report.html \
+		--plugin qa/check-charset.py,qa/check-fea.py,qa/check-googlesans.py,qa/check-sources.py \
+		--profile qa/googlesans-profile.toml --configuration qa/googlesans-config.toml \
+		sources/GoogleSansFlex.glyphspackage
```
RickyDaMa: Make it a separate recipe so it can be run standalone please. I don't think we want it hardcoded on every build, just on CI we would run source checks first.

RickyDaMa (commented):
Looking good, except I don't like this one part

RickyDaMa (approved):

